PR #72, opened 2020-09-17
Implement Kafka (compacted) topic-based persistence.

asavelyev01:


build.sbt
```diff
@@ -66,6 +67,21 @@ lazy val cassandra = (project in file("cassandra"))
     IntegrationTest / testFrameworks += new TestFramework("weaver.framework.TestFramework")
   )
 
+lazy val `kafka-persistence` = (project in file("kafka-persistence"))
+  .dependsOn(core)
+  .configs(IntegrationTest)
+  .settings(commonSettings)
+  .settings(
+    name := "kafka-flow-kafka-persistence",
```
rtar, line 75: Naming things is hard :)

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/package.scala
```diff
@@ -0,0 +1,52 @@
...
+
+package object kafkapersistence {
+
+  implicit class PartitionFlowOfCompanionOps(val self: PartitionFlowOf.type) extends AnyVal {
+
+    /**
+      * Creates PartitionFlowOf which on partition assignment reads respective partition of "snapshot" (usually compacted)
+      * topic and eagerly recovers all the state from it.
+      */
+    def eagerRecoveryKafkaPersistence[F[_] : Concurrent : Timer : Parallel : MeasureDuration : LogOf, S](
```
rtar, line 19: I see that formatting style is a bit different from the one used in Kafka Flow elsewhere. Should we try to keep it consistent / use scalafix rules or that is fine?

rtar, line 19: Why it should be an implicit extension method? Not saying it should not, just could not yet understand.

asavelyev01: In absence of scalafmt/scalafix I just used IntelliJ formatter. I'd be happy to use whatever tool to unify the format style.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,167 @@
...
+    S
+  ], S, A](stateTopic: Topic): KafkaPersistence[F, KafkaKey, S, A] = {
+    val snapshotsOf = new SnapshotsOf[F, KafkaKey, S] {
+      override def apply(key: KafkaKey): F[Snapshots[F, S]] =
+        Snapshots.of(
+          key,
+          new SnapshotDatabase[F, KafkaKey, S] {
+            override def get(key: KafkaKey) =
+              for {
+                state <- MonadState.get
```
rtar, line 78: Why use `MonadState` if you have `Sync` anyway? `Ref` could be easier and more performant option here.

rtar: Actually, when I look at the code, there is a point of `MonadState`, because there is a plenty of unit-testable logic here. But then why `Sync`?

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,167 @@
...
+
+final case class KafkaPersistence[F[_], K, S, A](
+                                                  keysOf: KeysOf[F, K],
+                                                  snapshots: SnapshotPersistenceOf[F, K, S, A],
+                                                  onRecoveryFinished: F[Unit]
+                                                )
+
+object KafkaPersistence {
+
+  private type Record = ConsumerRecord[String, ByteVector]
```
rtar, line 35: If we are depending on `kafka-journal` stuff anyway, for now, then it might make sense to just use `ConsRecord`, which is defined like following: `type ConsRecord = ConsumerRecord[String, ByteVector]`.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/package.scala
asavelyev01: My intention was to stress the fact that it's yet another constructor in `PartitionFlowOf` toolset and to provide consistent experience to client code writers, like `PartitionFlowOf.xxx` given the correct imports. Otherwise I'd need to come up with separate creatively named class or constructor object and as you rightly said, naming is hard:)

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,167 @@
...
+        for {
+          stateData <- readState(
+            consumerOf = consumerOf,
+            consumerConfig = consumerConfig,
+            stateTopic = stateTopic,
+            partition = partition
+          )
+          stateRef <- Ref.of(stateData)
+        } yield stateRef.runState { implicit state =>
+          KafkaPersistence[F, S, A](stateTopic)
```
rtar, line 62: Elsewhere, I pass state explicitly, and then use `stateInstance` method from the same library to pass it. I am not sure which is the best way, but we might want to agree on a single way to do that.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  private def readState[F[_] : BracketThrowable : FromBytes[*[_], String] : Log](
+                                                                                  consumerOf: ConsumerOf[F],
+                                                                                  consumerConfig: ConsumerConfig,
+                                                                                  stateTopic: Topic,
+                                                                                  partition: Partition
+                                                                                ): F[BytesByKey] = {
+    consumerOf
+      .apply[String, ByteVector](
+        GenLens[ConsumerConfig](_.common.clientId).modify(_.map(cid => s"$cid-state-$partition"))(consumerConfig)
```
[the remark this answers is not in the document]
asavelyev01: Not really, I can do that with `copy`

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: I need Sync here because I directly call `Snapshots.of` which asks `Sync`, and I probably shouldn't be doing it. Let me see if I can get rid of it.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Great point, done!

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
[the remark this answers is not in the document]
asavelyev01: @rtar dropped monocle, but TBH now I like it somewhat less :|

build.sbt
rtar: We, probably, should do something so `kafka-flow-cassandra` and `kafka-flow-kafka-persistence` have more similar names. I am having lack of fantasy today though :(

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,166 @@
...
+import com.olegpy.meow.effects._
+import scodec.bits.ByteVector
+
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+final case class KafkaPersistence[F[_], K, S, A](
+                                                  keysOf: KeysOf[F, K],
+                                                  snapshots: SnapshotPersistenceOf[F, K, S, A],
+                                                  onRecoveryFinished: F[Unit]
```
rtar, line 30: It seems that `onRecoveryFinished` does an important internal job in an implementation. Should it really be exposed in API?

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,166 @@
...
+          targetOffset <- BracketThrowable[F].fromOption(
+            endOffsets.get(statePartition),
+            MissingOffsetError(statePartition)
+          )
+          bytesByKey <- readPartition(targetOffset)
+        } yield bytesByKey
+      }
+  }
+
+  implicit def nelAsIterable[A]: NonEmptyList[A] => IterableOnce[A] = _.toList
```
rtar, line 164: Just call `toList` explicitly where it is needed?

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Well I can say that no members of `KafkaPersistence` are supposed to be used by the application code, it may (and should) be an opaque type for the client. Thanks for the reminder, will do that.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Dropped `Sync`,  used externally provided factory for state containers.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Made `monadState` passed explicitly.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Done.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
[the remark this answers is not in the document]
rtar: @asavelyev01 To be honest, I did not intend you to drop Monocle. I only thought `GenLens[ConsumerConfig](_.common.clientId)` might be a heavyweight operation and you might want to put it to some value, i.e. do something like this:
```
val lens = GenLens[ConsumerConfig]
```
To be completely honest, I do not even know if it is a heavyweight operation :blush:

rtar:
@asavelyev01 It looks like the build on 2.12.x fails with the following error (see "Checks" tab).

build.sbt
rtar: @asavelyev01 How about `kafka-flow-persistence-cassandra` and `kafka-flow-persistence-kafka`?

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,169 @@
...
+
+  private type BytesByKey = Map[String, ByteVector]
+
+  private object BytesByKey {
+    val zero: BytesByKey = Map.empty
+  }
+
+  private case class MissingOffsetError(topicPartition: TopicPartition) extends NoStackTrace
+
+  final case class Of[F[_], K, S, A](create: Partition => F[KafkaPersistence[F, K, S, A]])
```
rtar, line 44: In other places, in Kafka Flow, these `Of` factories are created as separate traits with separate files. I have no strong opinion on how it should be (I actually thinking if we can get rid of them altogether), but it might be cool to keep the style consistent for sake of readability.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,169 @@
...
+          },
+          buffer
+        )
+    }
+    val keysOf = new KeysOf[F, KafkaKey] {
+      override def apply(key: KafkaKey) = Keys.empty
+
+      override def all(applicationId: String, groupId: String, topicPartition: TopicPartition) =
+        Stream.lift(monadState.get
+          .map(bytes => Stream.from[F, List, KafkaKey](bytes.keys.map(KafkaKey(applicationId, groupId, topicPartition, _)).toList))).flatten
```
rtar, line 103: This line seems to be a bit intensive? Is it worth to split it to several pieces or lines?

build.sbt
asavelyev01: @rtar  I like the idea!

asavelyev01: I hope changing the name of already released module is not a big problem at this early stage.

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Done (with `toIterable`)

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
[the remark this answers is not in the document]
asavelyev01: Good point, it will be called once per each assigned partition. Changed to store once-generated lense in a static object.

build.sbt
asavelyev01: Renamed the modules

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Extracted `Of` as `KafkaPersistenceOf`

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: Agree, moved stream creation logic to a helper method.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
```diff
@@ -0,0 +1,59 @@
+package com.evolutiongaming.kafka.flow.kafkapersistence
```
rtar, line 1: How about `com.evolutiongaming.kafka.flow.persistence`, i.e. without `kafka`?

kafka-persistence/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/package.scala
asavelyev01: Added scalafmt but only formatted the new module so far

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistence.scala
asavelyev01: The thing is I have a package object which defines an extension constructor `PartitionFlowOf.eagerRecoveryKafkaPersistence`, so the client will of using this extension is to be expressed by the import `com.evolutiongaming.kafka.flow.kafkapersistence._`. With this regard I'd like to keep the package name distinct.

PR #125, opened 2020-12-17
Store only one snapshot per key

rtar:
This is required to avoid growing number of stored snapshots forever for long living entities. The change is backwards compatible and deployment without a downtime is supported.

### DIFF
--- a/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/CassandraSnapshots.scala
+++ b/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/CassandraSnapshots.scala
@@ -57,8 +57,8 @@ class CassandraSnapshots[F[_]: MonadThrowable: Clock, T](
     } yield ()
 
   def get(key: KafkaKey): F[Option[KafkaSnapshot[T]]] =
-    getLegacy(key) flatMap { snapshot =>
-      if (snapshot.isDefined) snapshot.pure else getNew(key)
+    getNew(key) flatMap { snapshot =>
+      if (snapshot.isDefined) snapshot.pure else getLegacy(key)
     }
 
   def getLegacy(key: KafkaKey): F[Option[KafkaSnapshot[T]]] = {
--- a/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/CassandraSnapshots.scala
+++ b/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/CassandraSnapshots.scala
@@ -27,18 +27,18 @@ class CassandraSnapshots[F[_]: MonadThrowable: Clock, T](
     for {
       preparedStatement <- session.prepare(
         """ UPDATE
-          |   snapshots
+          |   snapshots_v2
           | SET
           |   created = :created,
           |   metadata = :metadata,
-          |   value = :value
+          |   value = :value,
+          |   offset = :offset
           | WHERE
           |   application_id = :application_id
           |   AND group_id = :group_id
           |   AND topic = :topic
           |   AND partition = :partition
           |   AND key = :key
-          |   AND offset = :offset
         """.stripMargin
       )
       created <- Clock[F].instant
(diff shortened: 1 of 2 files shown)


persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/CassandraSnapshots.scala
```diff
@@ -56,7 +56,12 @@ class CassandraSnapshots[F[_]: MonadThrowable: Clock, T](
       _ <- session.execute(boundStatement).first.void
     } yield ()
 
-  def get(key: KafkaKey): F[Option[KafkaSnapshot[T]]] = {
+  def get(key: KafkaKey): F[Option[KafkaSnapshot[T]]] =
+    getLegacy(key) flatMap { snapshot =>
```
theide: it should be the other way round, getNew and fall back to getLegacy, or am I missing something here?

rtar: Yes, I fixed it :)

persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/SnapshotSchema.scala
```diff
@@ -32,6 +32,21 @@ private[snapshot] object SnapshotSchema {
...
+          |application_id TEXT,
+          |group_id TEXT,
+          |topic TEXT,
+          |partition INT,
+          |key TEXT,
+          |offset BIGINT,
+          |created TIMESTAMP,
+          |metadata TEXT,
+          |value BLOB,
+          |PRIMARY KEY(application_id, group_id, topic, partition, key)
```
asavelyev01, line 47: Seems like it will define partition key as `application_id`. Doesn't it mean that it will put all data on a single node per application?

rtar: If it was `PRIMARY KEY(application_id)`, it would, indeed put everything on a single node.

Or, to give another example, if you write `PRIMARY KEY(application_id, group_id, topic, partition)`, then everything related to the partition will be on a single node, because other fields in the key will stay the same. Of course, if the partition is the same, but `application_id` are different, then Cassandra will be able to put the data to several nodes for _different_ applications.

In our case, i.e. `PRIMARY KEY(application_id, group_id, topic, partition, key)`, will be able to use several nodes per application, group, topic, partition etc. because we have `key` there.

Everything related to one key will have to go to a single node though.

asavelyev01: From [Cassandra documentation](https://cassandra.apache.org/doc/latest/cql/ddl.html#partition-key):

> A CQL primary key is composed of 2 parts:

>- the partition key part. It is the first component of the primary key definition. It can be a single column or, using additional parenthesis, can be multiple columns. A table always have at least a partition key, the smallest possible table definition is:
`CREATE TABLE t (k text PRIMARY KEY);`
>- the clustering columns. Those are the columns after the first component of the primary key definition, and the order of those columns define the clustering order.

>Some example of primary key definition are:
`PRIMARY KEY (a)`: a is the partition key and there is no clustering columns.
`PRIMARY KEY (a, b, c)` : a is the partition key and b and c are the clustering columns.
`PRIMARY KEY ((a, b), c)` : a and b compose the partition key (this is often called a composite partition key) and c is the clustering column.
>Within a table, CQL defines the notion of a partition. A partition is simply the set of rows that share the same value for their partition key. 

Given the above, we should read `PRIMARY KEY(application_id, group_id, topic, partition, key)` as "`application_id` is the partition key and `group_id, topic, partition, key` is the sorting key, so we put rows with the same `application_id` in the same partition and on the same node".

rtar: Yes, you are right. Will fix it.

asavelyev01: Thanks @rtar 👍

rtar: @asavelyev01 It is such a pleasure when there is someone to review your code. I would have spent days debugging why everything went to one node suddenly if not your comment.

PR #130, opened 2020-12-21
fix multi topic subscription

dfahritdinov:


core/src/test/scala/com/evolutiongaming/kafka/flow/ConsumerFlowSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+import com.evolutiongaming.catshelper.LogOf
+import com.evolutiongaming.kafka.flow.kafka.Consumer
+import com.evolutiongaming.kafka.journal.ConsRecords
+import com.evolutiongaming.skafka.consumer.{ConsumerRecords, RebalanceListener}
+import com.evolutiongaming.skafka.{Offset, OffsetAndMetadata, Partition, Topic, TopicPartition}
+import munit.FunSuite
+
+import scala.concurrent.duration._
+
+class ConsumerFlowSpec extends FunSuite {
```
rtar, line 16: CI is complaining about imports in this file:
https://github.com/evolution-gaming/kafka-flow/pull/130/checks?check_run_id=1590071724

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/Consumer.scala
```diff
@@ -19,8 +20,11 @@ import scodec.bits.ByteVector
   */
 trait Consumer[F[_]] {
 
+  @deprecated("Use Consumer.subscribe with topics set parameter instead", "0.5.0")
```
rtar, line 23: If we are not using it in KafkaFlow code anymore, may be let's delete it?

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
```diff
@@ -132,4 +104,50 @@ object ConsumerFlow {
 
   }
 
+  private class RebalanceListenerImpl[F[_]: BracketThrowable: LogOf](
```
rtar, line 107: Create a new trait with implementation in companion object, so the unit could be tested in isolation and stubbed easily?

core/src/test/scala/com/evolutiongaming/kafka/flow/ConsumerFlowSpec.scala
dfahritdinov: Fixed

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/Consumer.scala
dfahritdinov: Yes, I'd like to drop the method too. But it will bread backward compatibility.

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
```diff
@@ -132,4 +104,50 @@ object ConsumerFlow {
...
+      foreachTopic(topicPartitions) { (topic, flow, partitions) =>
+        for {
+          log <- log[F]
+          _ <- log.prefixed(topic).info(s"$partitions lost, removing from topic flow")
+          _ <- flow.remove(partitions)
+        } yield ()
+      }
+
+    private def foreachTopic[A](topicPartitions: NonEmptySet[TopicPartition])(
+      f: (Topic, TopicFlow[F], NonEmptySet[Partition]) => F[A]
```
rtar, line 144: This signature looks a bit heavyweight... It also adds a cognitive load because `foreachTopic` is kind of a new DSL. :thinking: 

Should we make a method out of `topicPartitions.filter(_.topic == topic).map(_.partition).toNes` instead, so we can still call something like `traverse_` on a result?

Or I am nitpicking here?

core/src/test/scala/com/evolutiongaming/kafka/flow/ConsumerFlowSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+        Action.Apply()
+      )
+    )
+  }
+
+}
+
+object ConsumerFlowSpec {
+
+  type F[A] = SyncIO[A]
```
rtar, line 88: Could it be that you do not need `SyncIO` and `Ref` here as `ConsumerFlow` only depends on `BracketThrowable` and `LogOf`?

It looks to me that `StateT` would be more than enough and will actually make the test a bit more simple, but it may be just me :blush:

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
dfahritdinov: Done

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/Consumer.scala
dfahritdinov: Done

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
dfahritdinov: That's relevant complain, fixed

core/src/test/scala/com/evolutiongaming/kafka/flow/ConsumerFlowSpec.scala
rtar: Ah, okay. I mixed it up. You cannot easily replace `BracketThrowable` with `StateT` (you can only do it for `MonadThrowable`), but, as most of the logic is now in `RebalanceListener` does not require `BracketThrowable` and only requires `Monad`, it should be possible to cover it using `State` monad (as it does not throw any exceptions, `StateT` is not required).

core/src/main/scala/com/evolutiongaming/kafka/flow/RebalanceListener.scala
```diff
@@ -0,0 +1,62 @@
...
+import cats.syntax.all._
+import com.evolutiongaming.catshelper.{BracketThrowable, LogOf}
+import com.evolutiongaming.kafka.flow.ConsumerFlow.log
+import com.evolutiongaming.kafka.flow.kafka.Consumer
+import com.evolutiongaming.skafka.{Partition, Topic, TopicPartition}
+import com.evolutiongaming.skafka.consumer.{RebalanceListener => SRebalanceListener}
+
+object RebalanceListener {
+
+  def apply[F[_]: BracketThrowable: LogOf](
```
rtar, line 13: Here you can replace `BracketThrowable` with `Monad` now and write `State` based unit test.

core/src/main/scala/com/evolutiongaming/kafka/flow/RebalanceListener.scala
```diff
@@ -0,0 +1,62 @@
...
+        for {
+          log <- log[F]
+          _ <- log.prefixed(topic).info(s"$partitions lost, removing from topic flow")
+          _ <- flow.remove(partitions)
+        } yield ()
+      }
+
+    private def foreachTopic[A](
+      topicPartitions: NonEmptySet[TopicPartition]
+    ): List[(Topic, TopicFlow[F], NonEmptySet[Partition])] =
```
rtar, line 51: - The method name `foreachTopic` might not be ideal anymore, because the method does not have a function argument now.
- You dot need `private` modifier anymore as this is an anonymous implementation of `SRebalanceListener` now, so the method is not visible anymore.

core/src/main/scala/com/evolutiongaming/kafka/flow/RebalanceListener.scala
dfahritdinov: Done

core/src/main/scala/com/evolutiongaming/kafka/flow/RebalanceListener.scala
dfahritdinov: Renamed

core/src/test/scala/com/evolutiongaming/kafka/flow/ConsumerFlowSpec.scala
dfahritdinov: Done

core/src/test/scala/com/evolutiongaming/kafka/flow/RebalanceListenerSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    val consumer: Consumer[F] = new Consumer[F] {
+      def subscribe(topics: NonEmptySet[Topic], listener: SRebalanceListener[F]): F[Unit] = ().pure[F]
+      def poll(timeout: FiniteDuration): F[ConsRecords] = ConsRecords.empty.pure[F]
+      def commit(offsets: NonEmptyMap[TopicPartition, OffsetAndMetadata]): F[Unit] = ().pure[F]
+      def position(partition: TopicPartition): F[Offset] = offset.pure[F]
+    }
+
+    def flow(topic: String) = new TopicFlow[F] {
+      def apply(records: ConsRecords): F[Unit] = State modify (_ + Action.Apply(topic, records))
```
rtar, line 114: It does not look `Action.Apply` is used in this test. I would consider removing it and returning `().pure[F]` here.

core/src/test/scala/com/evolutiongaming/kafka/flow/RebalanceListenerSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+  val partition2 = Partition.unsafe(2)
+  val partition3 = Partition.unsafe(3)
+  val partition4 = Partition.unsafe(4)
+  val offset = Offset.unsafe(0)
+
+  type F[A] = State[Context, A]
+
+  implicit val logOf = LogOf.empty[F]
+
+  case class Context(actions: List[Action] = Nil) {
```
rtar, line 93: I feel the test is good enough and elegant enough already to keep, but I would consider using "stub-style" instead of "mock-style" here.

I.e. store `Map[Topic, (Partition, Offset)]` instead of `List[Action]` and update that map instead.

Feel free to ignore this comment, if you feel "mock-style" test is better in this case as I have no strong opinion. The test already looks very cool.

See also https://martinfowler.com/articles/mocksArentStubs.html for inspiration.

rtar:
@dfahritdinov I feel the commit is good to merge now. I wrote some comments, but these are really minor and feel free to ignore them. Feel free to press "Squash and merge" when you feel so.

core/src/test/scala/com/evolutiongaming/kafka/flow/KafkaFlowSpec.scala
```diff
@@ -29,83 +22,102 @@ class KafkaFlowSpec extends FunSuite {
...
+    assertEquals(
+      result,
+      State(
+        actions = List(
+          Action.ReleaseConsumer,
+          Action.ReleaseTopicFlow,
+          Action.Poll(consumerRecords(consumerRecord(partition = 0, offset = 0))),
+          Action.Poll(ConsumerRecords.empty),
+          Action.Poll(ConsumerRecords.empty),
+          Action.Subscribe(NonEmptySet.of(topic))(RebalanceListener.empty),
```
rtar, line 47: 2.12.x build is complaining about something here:
https://github.com/evolution-gaming/kafka-flow/pull/130/checks?check_run_id=1595806851

core/src/test/scala/com/evolutiongaming/kafka/flow/RebalanceListenerSpec.scala
dfahritdinov: Thanks for pointing me on the article. I don't see any reasons why the test cannot be converted to "stub-style".

Done.

core/src/test/scala/com/evolutiongaming/kafka/flow/RebalanceListenerSpec.scala
dfahritdinov: Done

core/src/test/scala/com/evolutiongaming/kafka/flow/KafkaFlowSpec.scala
dfahritdinov: Fixed

PR #135, opened 2021-01-06
use JsonCodec[Try] instead of JsonCodec[F] in JournalParser

dfahritdinov:
Motivation: 

Reduce pressure on end users from applying tagless final approach by dropping unnecessary type classes.

Constructing `JournalParser` requires implicit instances of both `JsonCodec[Try]` and `JsonCodec[F]`. 
To simplify method signature, one can be created from another:
```
// requires instance of ToTry[F]
implicit val jsonCodecTry = JsonCodec.summon[F] mapK ToTry.functionK[F]
```
or
```
// requires instance of ApplicativeError[F, *]
implicit val jsonCodecF = jsonCodec mapK FunctionK.liftFunction[Try, F](MonadThrowable[F].fromTry)
```

`ApplicativeError[F, *]` is supertype `MonadError[F, *]` witch is already in context. That makes approach of passing `JsonCodec[Try]` more preferable as instance of `JsonCodec[F]` can be created without additional constrains like `ToTry[F]`.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/journal/JournalParser.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/journal/JournalParser.scala
@@ -1,8 +1,8 @@
 package com.evolutiongaming.kafka.flow.journal
 
+import cats.arrow.FunctionK
 import cats.syntax.all._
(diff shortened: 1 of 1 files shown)


rtar:
Awesome!

PR #138, opened 2021-01-12
add empty PersistenceModule

dfahritdinov:


core/src/main/scala/com/evolutiongaming/kafka/flow/journal/JournalDatabase.scala
```diff
@@ -66,4 +66,11 @@ object JournalDatabase {
 
     }
 
+  def empty[F[_], K, R](implicit F: Applicative[F]): JournalDatabase[F, K, R] =
```
rtar, line 69: The following style is used elsewhere in `kafka-flow`. I do not mind another style, but we might want to keep it consistent:
```
def empty[F[_]: Applicative, K, R]: JournalDatabase[F, K, R] = ...
```
And then `().pure[F]` is used instead of `F.unit`.

core/src/main/scala/com/evolutiongaming/kafka/flow/key/KeyDatabase.scala
```diff
@@ -56,4 +56,11 @@ object KeyDatabase {
 
     }
 
+  def empty[F[_], K](implicit F: Applicative[F]): KeyDatabase[F, K] =
```
rtar, line 59: The same comment about the style of implicit `Applicative[F]` argument as above.

core/src/main/scala/com/evolutiongaming/kafka/flow/journal/JournalDatabase.scala
dfahritdinov: Redone in common for the project style

PR #154, opened 2021-01-29
Optimize Snapshot persistence in KafkaFlow Snapshot

dfahritdinov:


theide:
LGTM

core/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/Snapshots.scala
```diff
@@ -101,31 +102,23 @@ object Snapshots {
...
-        case Some(v) if v == version => true
-        case _       => false
-      }
+    def persisted(persistedValue: S): Snapshot[S] =
+      if (value == persistedValue) copy(persisted = true)
+      else self
   }
 
   object Snapshot {
     final case class Version(value: Long) {
```
rtar, line 117: Unused class

core/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/Snapshots.scala
```diff
@@ -61,20 +62,20 @@ object Snapshots {
...
 
     def flush = {
       for {
         snapshot <- buffer.get
         _ <- snapshot traverse_ { snapshot =>
-          if(!snapshot.persisted) {
+          if (!snapshot.persisted) {
             for {
               _ <- database.persist(key, snapshot.value)
               _ <- buffer.modify {
```
rtar, line 76: I think this method is not intended to be called concurrently anyway, so I suggest to replace `buffer.modify` by `buffer.set(snapshot.copy(persisted = true))` and simplify all the logic (i.e. throw out `def persisted(persistedValue: S)` method.

rtar:
@dfahritdinov I have added some random comments, but feel free to merge. It looks good to me too.

core/src/main/scala/com/evolutiongaming/kafka/flow/snapshot/Snapshots.scala
dfahritdinov: Here `MonadState` is backed by `Ref`:
```scala
  class RefMonadState[F[_]: Monad, S](ref: Ref[F, S]) extends MonadState[F, S] {
    val monad: Monad[F] = implicitly
    def get: F[S] = ref.get
    def set(s: S): F[Unit] = ref.set(s)
    def inspect[A](f: S => A): F[A] = ref.get.map(f)
    def modify(f: S => S): F[Unit] = ref.update(f)
  }
```
where method `update` is thread safe:
> Modifies the current value using the supplied update function. If another modification occurs between the time the current value is read and subsequently updated, the modification is retried using the new value. Hence, f may be invoked multiple times.

Using `set` instead of `modify` will bring race condition: `buffer.get` followed up by `buffer.set` are not atomic, while `buffer.modify` guarantee that operation will be retried if underline data changed. 
Together with idempotent implementation of `Snapshot#persisted` state always will be consistent.

rtar: This code is not meant to be called concurrently.

There is only one thread working with one key at a time by design. I.e. if you search for `MonadState` in `kafka-flow`, you will see plenty of "get and then set without using modify" style operations everywhere in the code.

It seems to me that trying to make them all consistent in concurrency will make the code really complicated for no reason. If we ever want to make concurrent operations on keys we might use something smarter.

There is only one potential reason for concurrent operations. I recently partially implemented "graceful shutdown" in `kafka-flow`, i.e. flushing state in `release` routine of `TimerFlowOf.unloadOrphaned`. This _may_ happen concurrently because `Resource` cancellation could be concurrent. It would be cool to analyze this code and, may be, introduce a semaphore there or something...

PR #163, opened 2021-02-16
Do not fail on malformed JSON but still fail on deserialisation to case class

dfahritdinov:
Ignore and log malformed JSON events, but still fail on deserialisation to case class.


rtar:
I think this logic does not belong to `kafka-flow` library. One can always provide more forgiving parser either as `JournalParser` or `Codec` without changing the library itself.

I also think that in most cases, one would want to actually fail the flow if JSON is malformed. If we could just ignore the random corrupt messages from the flow, we would need to care using persistence etc.

PR #165, opened 2021-02-17
Use skafka Consumer

dfahritdinov:
The PR should simplify codebase by dropping non necessary `Consumer`.
Also `groupId` now passed on `KafkaModule.of` level, making usage simpler.


core/src/test/scala/com/evolutiongaming/kafka/flow/RebalanceListenerSpec.scala
```diff
@@ -86,7 +87,7 @@ object RebalanceListenerSpec {
   val partition4 = Partition.unsafe(4)
   val offset = Offset.unsafe(0)
 
-  type F[A] = State[Context, A]
+  type F[A] = StateT[SyncIO, Context, A]
```
rtar, line 90: Why do we want to use `StateT[SyncIO, Context, *]` instead of `State[Context, *]` here? The code under testing, i.e. `RebalanceListener` only requires `Monad`.

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
```diff
@@ -0,0 +1,96 @@
...
+  OffsetAndTimestamp,
+  RebalanceListener => SRebalanceListener
+}
+import com.evolutiongaming.skafka._
+import scodec.bits.ByteVector
+
+import java.util.regex.Pattern
+import scala.concurrent.duration.FiniteDuration
+
+abstract class TestConsumer[F[_]](implicit F: ApplicativeError[F, Throwable]) extends Consumer[F, String, ByteVector] {
```
rtar, line 17: This is quite large class. What is the reason why we do not want to use simplified `Consumer`? It made the tests more simple and allowed us to very strictly control what methods did we call in a library.

It might be important, because calling some of the methods could cause deadlocks or not work at all.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
```diff
@@ -1,83 +1,80 @@
...
+  KafkaConfig,
+  KafkaHealthCheck,
+  RandomIdOf,
+  KafkaConsumerOf => JournalConsumerOf,
+  KafkaProducerOf => JournalProducerOf
+}
+import com.evolutiongaming.skafka.consumer._
+import com.evolutiongaming.skafka.producer.{Producer, ProducerMetrics, ProducerOf}
+import com.evolutiongaming.smetrics.{CollectorRegistry, MeasureDuration}
+import monocle.macros.syntax.lens._
```
rtar, line 19: Do we really want to introduce a dependency on monocle macros into the library code?

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
```diff
@@ -1,83 +1,80 @@
...
+  def producerOf: Resource[F, Producer[F]]
 
 }
 object KafkaModule {
 
   def of[F[_]: ConcurrentEffect: ContextShift: FromTry: ToTry: ToFuture: Timer: LogOf](
     applicationId: String,
-    config: ConsumerConfig,
+    groupId: String,
+    config: KafkaConfig,
```
rtar, line 35: It looks a bit ugly to pass the config and then change it in the code. The user might be surprised and annoyed when he or she tries to change part of the config, for it to only be changed in `KafkaModule` code. I know it was before your change.

What if we separate concerns and construct appropriate `KafkaConfig` in a separate class, and then just pass it here for further use?

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
```diff
@@ -50,7 +51,7 @@ object TopicFlowMetrics {
   implicit def topicFlowOfMetricsOf[F[_]: Monad: MeasureDuration]: MetricsOf[F, TopicFlowOf[F]] =
     topicFlowMetricsOf[F] transform { implicit metrics => topicFlowOf =>
       new TopicFlowOf[F] {
-        def apply(consumer: Consumer[F], topic: Topic) =
+        def apply(consumer: Consumer[F, String, ByteVector], topic: Topic) =
```
rtar, line 54: The need to drag `[String, ByteVector]` everywhere could be another reason why we might want to have our own `Consumer`.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
```diff
@@ -1,83 +1,80 @@
...
+  def consumerOf: Resource[F, Consumer[F, String, ByteVector]]
+  def producerOf: Resource[F, Producer[F]]
 
 }
 object KafkaModule {
 
   def of[F[_]: ConcurrentEffect: ContextShift: FromTry: ToTry: ToFuture: Timer: LogOf](
     applicationId: String,
-    config: ConsumerConfig,
+    groupId: String,
```
rtar, line 34: I think the reason that `ConsumerOf` was introduced is that one application might want to read one and the same topic several times for different purposes with different consumer group.

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
dfahritdinov: > It might be important, because calling some of the methods could cause deadlocks or not work at all.
All other methods will raise exception, don't see a way of getting dead lock.

The change makes production code more straightforward and simple, for tests we can define separate implementation.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: Why we should not?

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: > What if we separate concerns and construct appropriate KafkaConfig in a separate class, and then just pass it here for further use?

Nice idea, let me try to implement it.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
dfahritdinov: We can define `type Consumer[F[_]] = Consumer[F, String, ByteVector]` somewhere to avoid it. But I don't see an issue here, it's not more verbose that anything else.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: That's bit unexpected usage, could you provide an example?

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
rtar: Throwing errors in runtime instead of preventing the methods to be called in compile time seems scary.

May be you could give me a small code sample where using `Consumer[F, String, ByteVector]` from `skafka` is more straightforward and simple than using the simplified one, i.e. `Consumer[F]`, so I can also see it?

I did not write that wrapper, it was there in the code before me. I just feel it is a good idea, so there are at least two persons thinking it is a good idea, most probably :blush:

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
rtar: It will make library migration to Scala 3 harder eventually. We will also be forcing Monocle dependency on the users, while they might prefer using shapeless or Tofu lens instead.

The worst part, we will actually be using specific Monocle version and the library users might use a different one, which is not binary compatible with ours.

I would not care at all if dependency was added to end-user application, but libraries is a bit more sensitive about the unnecessary dependencies.

In this specific case, we can survive the pain of updating the case class manually.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
rtar: Defining two distinct `KafkaFlow.resource` resources in one application.

To give an example, one might want to read game journal topic both for sending DWH events and for generating game stats for purpose of showing these to players.

dfahritdinov: Then one should define two `KafkaModule` in the application that will run under the hood two Kafka consumers.

I think passing configuration somewhere after creation of `KafkaModule` is wrong approach that makes clients pass the configuration deeper throw multiple layers.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: Then I'll drop `monocle` dependency at runtime & use old good `.copy(...)`

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
dfahritdinov: > Throwing errors in runtime instead of preventing the methods to be called in compile time seems scary.

You are right that failing in runtime is evil, but hope we can make exception for tests as they always compile & run.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
rtar: So you are saying that writing `Consumer[F, String, ByteVector]` in 24 places is not more verbose than writing `Consumer[F]`?

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
rtar: In this case you will get two health checks etc., which seems to be a no-go.

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
rtar: Then I do not understand this statement:
> All other methods will raise exception, don't see a way of getting dead lock.

You said it will raise the exceptions while the will not on production.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: Why not? Two different Kafka clients should have independent health check as far as they can fail independently.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
dfahritdinov: I'm saying that using `type Consumer[F[_]] = Consumer[F, String, ByteVector]` makes it looks the same but without additional abstraction w/ logic, Of factory and misleading between skafka and flow consumers.

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
dfahritdinov: I don't get your last message, sorry.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
rtar: I think `Of` factory question is not anyhow related to this topic. It is a different discussion. You can have `Of` with `skafka` consumer or with our own with exactly the same success.

It is `Consumer[F]` vs `Consumer[F, String, ByteVector]`.

I see that the current pull request in this area:
(1) changes `Consumer[F]` to `Consumer[F, String, ByteVector]`,
(2) creates a large `TestConsumer` class with most methods failing,
(3) allows the one to accidentally call the much larger set of methods on underlying consumer.

Does it really simplify or make the code straightforward in this specific area? It creates a lot of lines of the code and the whole additional area to fail in. I agree `type` resolves (3), but it does not resolve other two issues.

It is a good programming practice to use as less power as we can. This is one of the reasons to use Tagless Final in the first place. Having our own very limited `Consumer` serves exactly this purpose - we limit ourselves to a subset of the functionality `skafka` provides getting more safety and simplifying the reasoning and the tests.

I agree about misleading name, i.e. the same as in `skafka`. I am not sure what the right policy could be. Naming things is hard. I think we were following naming convention of `kafka-journal`, i.e. to give exactly the same name for the things we want to be used instead. We could think about another policy and name it differently, i.e. `ByteConsumer` or something like that. I am not sure it is better. What do you think? :thinking:

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
rtar: We will have 3-4 clients then:
1. Client for consuming journals.
2. Client for producing messages for the applications where it is required.
3. Client for producing and consuming health checks.
4. Client for producing and consuming health checks.

So, (3) and (4) will do exactly the same, but somehow confirm the independence of (1) and (2), or I misunderstood the logic I see in the code?

core/src/test/scala/com/evolutiongaming/kafka/flow/TestConsumer.scala
rtar: Calling some of the methods in `com.evolutiongaming.skafka.Consumer` can cause a deadlock, because `skafka` tries to call all of them serially. See implementation of `skafka` for more details (https://github.com/evolution-gaming/skafka/blob/master/skafka/src/main/scala/com/evolutiongaming/skafka/consumer/Consumer.scala#L232).

Limiting ourselves to a small subset of `Consumer` functionality is a great way to avoid such a situation and also simplify testing and the code. It is a least-power principle. We only expose what we want to use.

dfahritdinov: > Calling some of the methods in com.evolutiongaming.skafka.Consumer can cause a deadlock, because skafka tries to call all of them serially. See implementation of skafka for more details (https://github.com/evolution-gaming/skafka/blob/master/skafka/src/main/scala/com/evolutiongaming/skafka/consumer/Consumer.scala#L232).

How come implementing **trait** can bring any of implementation details of skafka?

rtar: > How come implementing trait can bring any of implementation details of skafka?
It is similar to using `Monad` instead of `Sync` despite both being traits. `Sync` has more methods and we want to avoid exposing them to underlying logic.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
dfahritdinov: > (1) changes Consumer[F] to Consumer[F, String, ByteVector]

Please check latest commit, that's not true anymore.

> creates a large TestConsumer class with most methods failing

I prefer to have less production code & simple implementation with some additional test logic.

> allows the one to accidentally call the much larger set of methods on underlying consumer.

_one_ - kafka-flow library, probably? There are no user interaction w/ end user API as far as I see.

Additional abstraction makes already overcomplicated code even more complicated. The change has purpose to simplify codebase and make readability easier and cheaper to maintain.

I agree that test implementation is huge and we should come out with better solution (might be unpopular in FP world mocks?), but should not stop us from making production code simpler.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
dfahritdinov: That's why one should not run multiple Kafka consumers in one JVM. Instead one should subscribe on multiple topics from the same consumer in case of using same brokers. 

For rare cases of using different brokers having independent health-check doesn't looks like an issue.

metrics/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlowMetrics.scala
rtar: > Additional abstraction makes already overcomplicated code even more complicated. The change has purpose to simplify codebase and make readability easier and cheaper to maintain.
> ?
> I agree that test implementation is huge and we should come out with better solution (might be unpopular in FP world mocks?), but should not stop us from making production code simpler.

From my subjective opinion this change makes codebase more complicated and less readable :shrug:

This library, currently, uses traits extensively to simplify the testing. It also use a unit testing approach based on Tagless Final (https://medium.com/wix-engineering/functional-testing-with-tagless-final-50eeacf5df6). From my opinion such approach allows the one to write the code which is shorter and more reliably that the one based on mocks. I will be happy to teach you the one in person.

I do not think mocks are popular among anyone these days regardless if they are FP, OOP etc. developers. The people tend to prefer stubs for various reasons. I might be mistaken. See https://martinfowler.com/articles/mocksArentStubs.html for more details on the topic.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
rtar: > That's why one should not run multiple Kafka consumers in one JVM. Instead one should subscribe on multiple topics from the same consumer in case of using same brokers.
What if we want to use the same (or different?) topic, but different `groupId`? Create new JVM instance for every `groupId`? It does not sound like a great idea. It will decrease microservice density a lot for no specific reason...

I think we wandered away from the topic though. The reason I have added `KafkaModule` at all, because I was trying to move as much boilerplate as possible from the user to the library. The original library only had `ConsumerOf` and user had to configure it himself.

The whole purpose of `KafkaModule` was "consumer + health check". Should we just remove healthcheck from `kafka-flow` altogether and `KafkaModule` as a consequence? :thinking:

PR #172, opened 2021-02-22
remove unused method in KeyDatabase

nikitapecasa:
remove unused method in KeyDatabase: `def all(applicationId: String, groupId: String): Stream[F, K]`


persistence-cassandra/src/it/scala/com/evolutiongaming/kafka/flow/key/KeySpec.scala
```diff
@@ -50,7 +44,7 @@ class KeySpec(val globalResources: GlobalResources) extends CassandraSpec {
       session    = CassandraSessionStub.injectFailures(cassandra.session, failAfter)
       keys      <- CassandraKeys.withSchema(session, cassandra.sync)
       _ <- failAfter.set(1)
-      keys <- keys.all("KeySpec", "integration-tests-1").toList.attempt
+      keys <- keys.all("KeySpec", "integration-tests-1", TopicPartition("topic", Partition.min)).toList.attempt
```
nikitapecasa, line 47: the only change where I had to add something, but I was not able to run the spec locally with or without the changes.
getting error like https://issues.apache.org/jira/browse/CASSANDRA-16304

PR #229, opened 2021-07-20
Improve kafka persistence

daron666:


Z1kkurat:
Didn't even notice this already existed, thanks @daron666!
Short list of the changes in this PR:
- using new RebalanceConsumer1 from skafka
- some changes to core functionality by @nikitapecasa, I will take a look at them once more to ensure everything is in order and nothing more need to be done
- my rework of the current implementation of kafka-based snapshot persistence - the old one was reading the snapshots in the overriden `PartitionFlowOf`, now it reads snapshots with caching only once when `KeysOf.all` is invoked
- the large integration test mostly by @nikitapecasa ensuring that snapshot recovery works properly

nikitapecasa:
Also this PR brings
```
  /** Safeguards a TopicFlow's resource to address a race between processing of messages and release of the resource.
    *
    * It provides following safeties in case of a fiber cancellation and resource's release:
    *
    *  - processing of current kafka messages won't be cancelled (TopicFlow.apply)
    *  - pending offsets would be committed
    *  - accumulated state would be persisted
    *  - resource would be released only after previous steps are completed
    *  - if resource is released before start of messages' processing, then it does nothing (no message processing, no commits, no persists)
    */
```

nikitapecasa:
Also during working on this PR it was discovered that current graceful shutdown sequence can be improved https://github.com/evolution-gaming/kafka-flow/issues/256

PR #246, opened 2021-09-03
Bump kafka-journal and other dependencies, migrate to new RebalanceCallback1

Z1kkurat:


Z1kkurat:
Latest build failure caused by https://github.com/embeddedkafka/embedded-kafka/issues/202.  
Will update scala 2.12 version to 2.12.14, should be OK with the currently pulled kafka version.

Z1kkurat:
image

core/src/main/scala/com/evolutiongaming/kafka/flow/RebalanceListener.scala
```diff
@@ -5,46 +5,50 @@ import cats.data.NonEmptySet
...
-  ): SRebalanceListener[F] = new SRebalanceListener[F] {
+  ): SRebalanceListener1[F] = new RebalanceListener1WithConsumer[F] {
 
-    def onPartitionsAssigned(topicPartitions: NonEmptySet[TopicPartition]) =
+    override def onPartitionsAssigned(topicPartitions: NonEmptySet[TopicPartition]): RebalanceCallback[F, Unit] =
       groupByTopic(topicPartitions) traverse_ { case (topic, flow, partitions) =>
         for {
-          log <- log[F]
-          _ <- log.prefixed(topic).info(s"$partitions assigned")
+          log <- log[F].lift
```
t3hnar, line 26: please minimise number of `lift` calls

t3hnar: like one before  `consumer.position` and one after

Z1kkurat: done

PR #325, opened 2022-03-19
Add filtering capabilities to PartitionFlow

Z1kkurat:
Micro-PoC that this filtering code works:
```scala
import cats.data.NonEmptyList
import cats.effect.IO
import cats.instances.all._
import cats.syntax.all._

import scala.collection.SortedMap
import scala.concurrent.ExecutionContext

implicit val cs = IO.contextShift(ExecutionContext.global)

val filterFun = (x: Int) => IO(x != 1)

val map1 =
  SortedMap("key1" -> NonEmptyList.of(1, 2, 3), "key2" -> NonEmptyList.of(1, 1, 1))
val map2 =
  SortedMap("key1" -> NonEmptyList.of(1, 1, 1))
val map3 =
  SortedMap("key1" -> NonEmptyList.of(1, 2, 3))

def filter(map: SortedMap[String, NonEmptyList[Int]]): List[(String, NonEmptyList[Int])] =
  map.toList
    .parTraverseFilter { case (key, records) =>
      records.toList.filterA(filterFun).map(filtered => NonEmptyList.fromList(filtered).map(nel => (key, nel)))
    }
    .unsafeRunSync()

filter(map1)
// val res0: List[(String, cats.data.NonEmptyList[Int])] = List((key1,NonEmptyList(2, 3)))

filter(map2)
// val res1: List[(String, cats.data.NonEmptyList[Int])] = List()

filter(map3)
// val res2: List[(String, cats.data.NonEmptyList[Int])] = List((key1,NonEmptyList(2, 3)))
```

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
@@ -1,10 +1,10 @@
 package com.evolutiongaming.kafka.flow
 
-import cats.Parallel
 import cats.data.NonEmptyList
 import cats.effect.concurrent.Ref
 import cats.effect.{Clock, Concurrent, Resource}
 import cats.syntax.all._
+import cats.{Applicative, Parallel}
 import com.evolutiongaming.catshelper.ClockHelper._
 import com.evolutiongaming.catshelper.{Log, LogOf}
 import com.evolutiongaming.kafka.flow.kafka.OffsetToCommit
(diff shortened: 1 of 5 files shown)


Z1kkurat:
Maybe it would be better to pass the filtering function as `Option[ConsRecord => F[Boolean]]` to be able to pass just `None` as default and don't do redundant work on collections when no filtering is needed 🤔

theide:
> Maybe it would be better to pass the filtering function as `Option[ConsRecord => F[Boolean]]` to be able to pass just `None` as default and don't do redundant work on collections when no filtering is needed 🤔

I think it's a good idea to not traverse everything if we can avoid it, we could also pass the whole `FilterRecord` as an option and only apply it if it's `Some`

theide (approved):

PR #332, opened 2022-04-06
Add read write config support for cassandra persistence

z0s0:
Currently there is not way to tune consistency level for read/write queries in cassandra-persistence as that level that is provided in `CassandraConfig.client` will be applied for both. My case is to increase consistency level for writes to EACH_QUORUM which will cause a huge performance penalty for reads.
Suggested fix 

- take ConsistencyConfig from kafka-journal
- add new factory method CassandraPersistence.withSchema(session, sync, consistencyConfig) which does the same as `withSchema(session, sync)` but also adds `setConsistencyLevel` on the level of `PreparedStatement` construction.
- move `PreparedStatement` construction to a separate companion as the queries to be referenced from 2 places.

Sample usage on the application side
```
evolutiongaming.appname.cassandra {
  consistency-config {
    read  = ${?CASSANDRA_READ_CONSISTENCY_LEVEL}
    write = ${?CASSANDRA_WRITE_CONSISTENCY_LEVEL}
  }
}
```

and just replace 
```
CassandraPersistence.withSchema(session, sync) to CassandraPersistence.withSchema(session, sync, consistencyConfig)
```

### DIFF
--- a/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/CassandraKeys.scala
+++ b/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/CassandraKeys.scala
@@ -18,6 +18,7 @@ import com.evolutiongaming.scassandra.syntax._
 import com.evolutiongaming.skafka.Partition
 import com.evolutiongaming.skafka.TopicPartition
 import com.evolutiongaming.sstream.Stream
+import CassandraKeys.Statements
 
 import java.time.LocalDate
 import java.time.ZoneOffset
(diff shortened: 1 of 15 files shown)


persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/CassandraKeys.scala
```diff
@@ -178,9 +93,181 @@ object CassandraKeys {
...
+    ): Stream[F, KafkaKey] = {
+      val boundStatement = Statements
+        .all(session, applicationId, groupId, segment, topicPartition)
+        .map(_.setConsistencyLevel(readConsistency))
+
+      Stream.lift(boundStatement).flatMap(session.execute).map { row =>
+        KafkaKey(
+          applicationId = applicationId,
+          groupId = groupId,
+          topicPartition = TopicPartition(
```
Z1kkurat, line 138: Do we need to decode it here if we have it as a parameter already? We also have this decoding at line 75, maybe we can extract it and reuse?

Z1kkurat (commented):
I'm thinking if we can unify this somehow to maybe not have two implementations that differ only in setting a consistency level. Maybe pass these levels as optional values to `CassandraJournals` class and pass an optional `ConsistencyConfig` to `CassandraJournals.withSchema`?  
Overall -- great job, thank you!

Z1kkurat (approved):
Great job, thank you!  
Let's also get an approve from @theide

theide (approved):
nice job, thanks!

PR #337, opened 2022-04-19
Upgrade to cats-effect 3

Z1kkurat:
This brings in cats-effect 3 and all the updated libraries. 
The library version now starts with `1.x.x`. 
The "CE2" version is branched from the current `master` in `series/0.x` branch. All updates to it should not make the version higher than `0.x.x` as `1.x.x` and further is reserved for CE3 versions in the `master` branch.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaFlow.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaFlow.scala
@@ -1,19 +1,17 @@
 package com.evolutiongaming.kafka.flow
 
-import cats.effect.Concurrent
-import cats.effect.Resource
-import cats.effect.Timer
+import cats.effect.{Concurrent, Resource}
 import cats.effect.implicits._
+import cats.effect.kernel.Outcome
 import cats.syntax.all._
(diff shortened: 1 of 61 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaFlow.scala
```diff
@@ -82,8 +78,12 @@ object KafkaFlow {
...
   def resource[F[_]: Concurrent: Retry](
     consumer: Resource[F, Consumer[F]],
-    flowOf: ConsumerFlowOf[F],
+    flowOf: ConsumerFlowOf[F]
   ): Resource[F, F[Unit]] =
-    stream(consumer, flowOf).drain.background
+    stream(consumer, flowOf).drain.background.map(_.flatMap {
+      case Outcome.Succeeded(fa) => fa
+      case Outcome.Errored(e)    => e.raiseError[F, Unit]
+      case Outcome.Canceled()    => new CancellationException().raiseError[F, Unit]
```
Phill101, line 86: ?
```scala
Concurrent[F].canceled
```

Phill101 (approved):
lgtm

theide (approved):

PR #351, opened 2022-06-03
Make Segments in CassandraKeys configurable (CE2)

Z1kkurat:


persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/CassandraKeys.scala
```diff
@@ -26,26 +26,27 @@ import java.time.ZoneOffset
 
 class CassandraKeys[F[_]: Monad: Fail: Clock](
   session: CassandraSession[F],
-  consistencyOverrides: ConsistencyOverrides = ConsistencyOverrides.none
+  consistencyOverrides: ConsistencyOverrides = ConsistencyOverrides.none,
+  segments: Segments
```
vilunov, line 30: FYI this constructor is public, therefore this is a breaking change. I recommend adding a default arg `segments: Segments = CassandraKeys.DefaultSegments,`, or introducing a deprecated `def this`

Z1kkurat: While I agree with your point, I honestly don't think anyone is using it now (especially via a constructor) so we're pretty safe in terms of backward compatibility. No users means no compatibility broken 😄  
Moreover, I wanted to make this constructor argument explicit without a default value because it can have a severe and unexpected impact on performance. Maybe it would be even better to make this constructor `private` and leave only factory methods in the companion

vilunov: let's make it private then?

Z1kkurat: I decided to go with new constructor methods to make users always provide `Segments` while one of the constructors defaults `consistencyOverrides` to `none`. So if someone was using `new CassandraKeys(session)` before, they will now get a deprecation error and can use `new CassandraKeys(session, segments)` while the old one is still there so it can't cause a binary incompatibility

theide (approved):

PR #371, opened 2022-08-19
Update dependencies

Z1kkurat:


Z1kkurat:
There's still a discrepancy between kafka-clients versions brought by kafka-launcher:0.0.12/skafka:12.2.2 (2.8.1) and kafka-journal:1.0.2 (2.7.2). I'll wait skafka:14.0.0 and a corresponding update to kafka-journal and bump a major version of this library so that kafka version is aligned.

PR #374, opened 2022-09-02
Fix table name in SnapshotSchema

pizzaeueu:


Z1kkurat:
Good catch @pizzaeueu! Thanks!

PR #385, opened 2022-09-20
Entity registry

Z1kkurat:
This PR adds an observability API to inspect the current state of all in-memory entities in form of `EntityRegistry`.  
An instance of such a registry is supposed to be created by the user and passed to one of `KeyStateOf` factory methods. When creating a new in-memory representation of an entity the framework will use the registry to associate the key with the function to obtain the current state for that key - in our case that's reading from a `Ref` where the state is stored. This association is removed when the entity is deleted from memory (when Fold or Tick return None as the new state).  

The registry can then be passed to any user-side code such as HTTP servers which can invoke its `getAll` method and obtain a full Map of currently stored states. `EntityRegistry#register` is not supposed to be called by user but is left public so that framework users could override it in case they want to implement it their own way.

By default we provide the following implementations: `empty` (no-op), `const` (using a constant immutable Map with a no-op registration) and `memory` (using `Ref` to store `K -> F[Option[S]]` pairs in `Map`). All of the classes are final as I'd like to prevent piggybacking on some internal APIs of the framework.


core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
```diff
@@ -0,0 +1,87 @@
...
+  */
+trait EntityRegistry[F[_], K, S] {
+
+  /** Registers an association between an entity's key and the function to obtain current state of it.
+    * This is used internally by the library and shouldn't be called by the library users.
+    * @param key entity key
+    * @param state computation (effectively function) to obtain the current value of the entity
+    * @return
+    */
+  def register(key: K, state: F[Option[S]]): Resource[F, Unit]
```
vilunov, line 28: why is this a resource? you add to scaladoc that freeing this resource removes the entity from registry

core/src/main/scala/com/evolutiongaming/kafka/flow/KeyFlow.scala
```diff
@@ -17,72 +20,84 @@ object KeyFlow {
 
   /** Create flow which persists snapshots, events and restores state if needed */
   def of[F[_]: Monad: Ref.Make: KeyContext, S, A](
+    key: KafkaKey,
```
vilunov, line 23: a bunch of broken apis, will you bump the major version? you could keep the original methods as deprecated and call the new ones with a fake instance of `KafkaKey`

core/src/main/scala/com/evolutiongaming/kafka/flow/KeyFlowOf.scala
```diff
@@ -3,15 +3,18 @@ package com.evolutiongaming.kafka.flow
 import cats.Monad
 import cats.effect.{Ref, Resource}
 import com.evolutiongaming.kafka.flow.persistence.Persistence
+import com.evolutiongaming.kafka.flow.registry.EntityRegistry
 import com.evolutiongaming.kafka.flow.timer.{TimerContext, TimerFlowOf}
 
 trait KeyFlowOf[F[_], S, A] {
 
   def apply(
+    key: KafkaKey,
```
vilunov, line 12: This one is a bit trickier. It's probably possible to make something like this:
```scala
@deprecated
trait KeyFlowOf[F[_], S, A] extends KeyFlowOf2[F, S, A] {

  def apply(
    context: KeyContext[F],
    persistence: Persistence[F, S, A],
    timers: TimerContext[F],
    additionalPersist: AdditionalStatePersist[F, S, A],
  ): Resource[F, KeyFlow[F, A]]

  final def apply(
    key: KafkaKey,
    context: KeyContext[F],
    persistence: Persistence[F, S, A],
    timers: TimerContext[F],
    additionalPersist: AdditionalStatePersist[F, S, A],
    registry: EntityRegistry[F, KafkaKey, S],
  ): Resource[F, KeyFlow[F, A]] = apply(fakeKafkaKey, context, persistence, timers, additionalPersist, EntityRegistry.noop)

}

trait KeyFlowOf2[F[_], S, A] {

  def apply(
    key: KafkaKey,
    context: KeyContext[F],
    persistence: Persistence[F, S, A],
    timers: TimerContext[F],
    additionalPersist: AdditionalStatePersist[F, S, A],
    registry: EntityRegistry[F, KafkaKey, S],
  ): Resource[F, KeyFlow[F, A]]

}
```

and use KeyFlowOf2 in the library code

core/src/main/scala/com/evolutiongaming/kafka/flow/KeyFlow.scala
Z1kkurat: Yes, I planned to increase the major version. This one, however, is an internal API, not used by the end user, so it's fine to change this. There are a lot of them in `KeyStateOf` along with new versions of methods since that's a part of public API

core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
Z1kkurat: It says __When the entity is evicted from memory, this association is also removed.__ Did you mean something else?

Z1kkurat: >this means that the association is removed when the entity is garbage collected

This is not what's meant actually. It means that it's going to be removed when we remove the entity from the internal cache. I'll rephrase this.

core/src/main/scala/com/evolutiongaming/kafka/flow/KeyFlow.scala
Z1kkurat: That makes sense. I'll remove deprecated methods but leave factory methods public in case anyone wants to re-use them while implementing custom instances.

Z1kkurat: Done.

core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
Z1kkurat: Done, please see if it's phrased clearer now

vilunov:
LGTM

core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
```diff
@@ -0,0 +1,87 @@
...
+  /** Immutable registry with a no-op `register` method. `get` and `getAll` query the passed map */
+  final class ConstEntityRegistry[F[_], K, S](values: Map[K, S])(implicit F: Applicative[F])
+      extends EntityRegistry[F, K, S] {
+    override def register(key: K, state: F[Option[S]]): Resource[F, Unit] = Resource.unit
+    override def get(key: K): F[Option[S]] = F.pure(values.get(key))
+    override def getAll: F[Map[K, S]] = F.pure(values)
+  }
+
+  /** Mutable registry that keeps the state in a `Ref` */
+  final class InMemoryEntityRegistry[F[_], K, S](ref: Ref[F, Map[K, F[Option[S]]]])(implicit F: Monad[F])
```
t3hnar, line 67: single `Ref` covering all entities - could be a bottleneck.

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
Z1kkurat: Maybe we could first see how it does and if it's not enough then we can review our options (premature optimization and all that).  
What would you recommend as an alternative, would `com.evolutiongaming.scache.Cache#loading` be a good choice here in your opinion given how it's partitioned? I could introduce a separate implementation based on it next to this one so that users can choose themselves and not be bound to `Cache`'s implementations/possible bugs in any way.

t3hnar:
```
partitions: Partitions[Ref[Map[K, S]]
```


In this case I disagree about "premature optimization and all that" argument :)

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/kafka/flow/registry/EntityRegistry.scala
```diff
@@ -0,0 +1,112 @@
...
+  ): F[EntityRegistry[F, K, S]] = {
+    val makeRef = Ref[F].of(Map.empty[K, F[Option[S]]])
+    Partitions.of(nrOfPartitions, _ => makeRef).map(partitions => new InMemoryEntityRegistry[F, K, S](partitions))
+  }
+
+  def const[F[_]: Applicative, K, S](values: Map[K, S]): EntityRegistry[F, K, S] =
+    new ConstEntityRegistry[F, K, S](values)
+
+  /** No-op registry, always returning None and an empty map on `get` and no-op on `register` */
+  final class EmptyEntityRegistry[F[_], K, S](implicit F: Applicative[F]) extends EntityRegistry[F, K, S] {
```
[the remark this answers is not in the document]
Z1kkurat: Done

PR #386, opened 2022-09-26
Entity registry [CE2]

Z1kkurat:
This is a backport of https://github.com/evolution-gaming/kafka-flow/pull/385

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaKey.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaKey.scala
@@ -11,5 +11,5 @@ final case class KafkaKey(
 )
 
 object KafkaKey {
-  implicit val hash: Hash[KafkaKey] = Hash.fromUniversalHashCode
+  implicit val hashKafkaKey: Hash[KafkaKey] = Hash.fromUniversalHashCode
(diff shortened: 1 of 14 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaKey.scala
```diff
@@ -1,10 +1,15 @@
...
   applicationId: String,
   groupId: String,
   topicPartition: TopicPartition,
   key: String
-)
\ No newline at end of file
+)
+
+object KafkaKey {
+  implicit val hash: Hash[KafkaKey] = Hash.fromUniversalHashCode
```
t3hnar, line 14: I propose to add `KafkaKey` suffix: `hashKafkaKey`, as sometimes you may have weird compile problems because of conflicting types with same name

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/kafka/flow/KafkaKey.scala
Z1kkurat: Makes sense, done

t3hnar (approved):

PR #387, opened 2022-10-12
Toggleable compression for application state

Z1kkurat:
This adds a built-in compression support for enriching user-defined instances of `ToBytes` and `FromBytes` with it.  
It also provides a way to report compression metrics to `CollectorRegistry[F]`

Usage example:
```scala
import cats.effect.IO
import cats.effect.syntax.resource._
import com.evolutiongaming.kafka.flow.FlowMetrics
import com.evolutiongaming.kafka.flow.metrics.syntax._
import com.evolutiongaming.kafka.flow.persistence.compression.{Compressor, Header}
import com.evolutiongaming.skafka.{FromBytes, ToBytes}
import com.evolutiongaming.smetrics.CollectorRegistry
import com.evolutiongaming.kafka.flow.persistence.compression.CompressorSyntax._
import scodec.bits.BitVector
import scodec.codecs.{bool, int32}

val registry: CollectorRegistry[IO] = CollectorRegistry.empty[IO]

// Application's state
final case class State(int: Int)
// Encoder of the application's state
val toBytes: ToBytes[IO, State] = (state, _) =>
  IO.fromTry(int32.encode(state.int).map(_.toByteArray).toTry)
// Encoder/decoder of metainformation header
implicit val headerToBytes: ToBytes[IO, Header] =
  (header, _) => IO.fromTry(bool.encode(header.compressed).map(_.toByteArray).toTry)
implicit val headerFromBytes: FromBytes[IO, Header] =
  (bytes, _) => IO.fromTry(bool.decode(BitVector(bytes)).map(result => Header(result.value)).toTry)
  
// Resulting instance can be passed to other parts of kafka-flow's API
for {
  flowMetrics <- FlowMetrics.of(registry)
  compressor <- 
    Compressor
      .of[IO](compressionThreshold = 10000)
      .map(_.withMetrics(flowMetrics.compressorMetrics("component")))
      .toResource
  toBytesWithCompression = toBytes.withCompression(compressor)
} yield ()
```


core/src/main/scala/com/evolutiongaming/kafka/flow/persistence/compression/Compressor.scala
```diff
@@ -0,0 +1,99 @@
...
+        def to(payload: ByteVector): F[ByteVector] = {
+          for {
+            result <-
+              if (payload.length >= compressionThreshold) {
+                compression.compress(payload).map((true, _))
+              } else {
+                (false, payload).pure[F]
+              }
+            (compressed, compressedBytes) = result
+            headerBytes <- headerToBytes.apply(Header(compressed = compressed), topic = "")
```
theide, line 74: why do we have topic here?

Z1kkurat: It's a part of `ToBytes` API from skafka, it's not going to affect anything here

core/src/main/scala/com/evolutiongaming/kafka/flow/persistence/compression/Compressor.scala
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        def from(bytes: ByteVector): F[ByteVector] = {
+          if (bytes.startsWith(uncompressedJsonStartBytes)) {
+            bytes.pure[F]
+          } else {
+            for {
+              result <- HeaderAndPayload.fromBytes(bytes)
+              (headerBytes, payload) = result
+              header <- headerFromBytes.apply(headerBytes.toArray, topic = "")
```
theide, line 86: same question as above, why do we have topic here

core/src/main/scala/com/evolutiongaming/kafka/flow/persistence/compression/Compressor.scala
theide: thank you!

PR #420, opened 2023-01-12
Add debug logging

edubrovski:


core/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlow.scala
```diff
@@ -80,6 +80,9 @@ object TopicFlow {
       def apply(records: ConsRecords) = {
         for {
           partitions <- cache.values
+          _ <- Log[F].debug(
+            s"got ConsRecords with partitions ${records.values.keySet}; cached partitions: ${cache.values}"
```
Z1kkurat, line 84: I think `cache.values` returns an `F[...]` here so there's not much sense in logging it. On line 82 you already have `partitions` and it's a `Map[Partition, PartitionFlow[F]]`. Probably you'll need only keys from that.  

We could also print information about received consumer records in some pretty way, let's say, "${partition} - ${offset of the last message for it}". Since `ConsumerRecords` is a wrapper around `Map[TopicPartition, Nel[ConsumerRecord[K, V]]]`, that shouldn't be an issue, I suppose. Please don't forget to create this string only inside `Log[F].debug {}` block so that we don't allocate a redundant String when debug logging is disabled

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
```diff
@@ -81,21 +81,26 @@ object ConsumerFlow {
...
-    def poll = {
+    def poll(logger: Log[F]) = {
       val flowList = flows.toList // optimization, execute toList once instead of on each `consumer.poll`
-      consumer.poll(config.pollTimeout) flatTap { consumerRecords =>
-        flowList traverse { case (topic, flow) =>
-          val topicRecords = consumerRecords.values filter { case (partition, _) =>
-            partition.topic == topic
-          }
-          flow(ConsumerRecords(topicRecords))
+      logger.debug(s"consumer.poll(${config.pollTimeout})") *>
```
Z1kkurat, line 86: Let's make it a for-comprehension, please. It's quite hard to read, especially, during the review (see the screenshot). 
Screenshot 2023-01-16 at 1 30 49 PM
I suggest transforming it to something like this:
```scala
for {
  _ <- logger.debug(s"consumer.poll(${config.pollTimeout})")
  consumerRecords <- consumer.poll(config.pollTimeout)
  _ <- logger.debug(s"polled records: ${ConsumerRecords.summaryShow.show(consumerRecords)}")
  _ <- flowList.traverse { case (topic, flow) =>
    val topicRecords = consumerRecords.values filter { case (partition, _) =>
      partition.topic == topic
    }
    flow(ConsumerRecords(topicRecords))
  }
  _ <- logger.debug(s"poll completed")
} yield consumerRecords
```
This one looks more like straightforward imperative code with fewer levels of nesting.

core/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlow.scala
```diff
@@ -78,8 +79,12 @@ object TopicFlow {
     val acquire = new TopicFlow[F] {
 
       def apply(records: ConsRecords) = {
+
         for {
           partitions <- cache.values
+          _ <- Log[F].debug(
```
Z1kkurat, line 85: Doesn't it duplicate what's going to be logged in `ConsumerFlow`?

edubrovski: Removed duplicate logging in `ConsumerFlow`

PR #422, opened 2023-01-19
Update debug logging

edubrovski:


Z1kkurat (approved):

edubrovski (commented):
this is a duplicate of what we already have in logs:
`DEBUG io-compute-blocker-2 c.e.k.f.k.Consumer poll timeout: 10 milliseconds in 0ms, result: journal.Player-4:4783140956..4783140959 records: 4, journal.Player-5:4797222859..4797222863 records: 5`

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
```diff
@@ -84,7 +84,6 @@ object ConsumerFlow {
     def poll(logger: Log[F]) = {
       val flowList = flows.toList // optimization, execute toList once instead of on each `consumer.poll`
       for {
-        _ <- logger.debug(s"consumer.poll(${config.pollTimeout})")
```
edubrovski, line 87: This is a duplicate of what we already had:
`DEBUG io-compute-blocker-2 c.e.k.f.k.Consumer poll timeout: 10 milliseconds in 0ms, result: journal.Player-4:4783140956..4783140959 records: 4, journal.Player-5:4797222859..4797222863 records: 5`

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -135,7 +132,7 @@ object PartitionFlow {
     } yield ()
 
     def processRecords(records: NonEmptyList[ConsRecord]) = for {
-      _ <- Log[F].debug(s"processing ${records.size} records from partition ${topicPartition.partition}")
```
edubrovski, line 138: somehow prefixes with topics and partitions are computed, so we don't need this

PR #427, opened 2023-01-25
Refactoring: remove usage of MTL

edubrovski:


stasimus:
@edubrovski should I close this MR?

PR #430, opened 2023-01-26
Parallelize processing of Kafka record on topic level

vilunov:


core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
```diff
@@ -84,11 +84,10 @@ object ConsumerFlow {
     def poll = {
       val flowList = flows.toList // optimization, execute toList once instead of on each `consumer.poll`
       consumer.poll(config.pollTimeout) flatTap { consumerRecords =>
-        flowList traverse { case (topic, flow) =>
-          val topicRecords = consumerRecords.values filter { case (partition, _) =>
-            partition.topic == topic
-          }
-          flow(ConsumerRecords(topicRecords))
+        val recordsByTopic = consumerRecords.values.groupBy(_._1.topic)
```
t3hnar, line 87:
```
consumerRecords
.values
.groupBy(_._1.topic)
.parFoldMapA { (topic, records) +> } 
```

please, we do not need use `flowList` here, right?

core/src/test/scala/com/evolutiongaming/kafka/flow/KafkaFlowSpec.scala
```diff
@@ -143,6 +145,15 @@ object KafkaFlowSpec {
 
   type F[A] = SyncIO[A]
 
+  // TODO rewrite in IO
+  implicit val parallelForSyncIO = new Parallel[F] {
```
t3hnar, line 149: there is `Parallel.identity` instead

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/kafka/flow/ConsumerFlow.scala
vilunov: we still need to call `apply` on flows even when we haven't received record for the topic, as we need to trigger timers for it

t3hnar: than I propose the only change to be done: replace `traverse` with `parFoldMapA`, you'd better leave `filter` inside of parallelised snippet

vilunov: the filter inside is not going to be parallelised, only the IO will be. the function itself is called sequentially

t3hnar: it's more complex than that, you can make a test to check it out.

t3hnar (approved):

PR #433, opened 2023-01-30
Call KeyContext.holding with timeout

edubrovski:


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -201,6 +203,12 @@ object PartitionFlow {
       _ <- Log[F].debug("done triggering timers")
     } yield ()
 
+    def getOffset(partitionKey: PartitionKey[F]): F[Option[Offset]] =
+      GenSpawn[F].race(Temporal[F].sleep(5.minutes), partitionKey.context.holding).flatMap {
```
Z1kkurat, line 207: There's a `def timeoutTo[A](fa: F[A], duration: Duration, fallback: F[A]): F[A]` in `GenTemporal`.  
```scala
import cats.effect.syntax.all._
partitionKey.context.holding.timeoutTo(5.minutes, Log[F].error(...))
```
Why 5 minutes, specifically? It's equal to the default Kafka's `max.poll.interval.ms`, having the timeout equal to that will cause it to behave less predictably as it's going to be randomly either time out before kicked out of the group or after that

edubrovski: Changed to 1 minute. I chose 1 minute because it's long enough to make sure the call hanged but it's less than 5 minutes (consumer poll timeout).

Z1kkurat (approved):

PR #435, opened 2023-01-31
Replace parTraverse with parTraverseN in PartitionFlow

edubrovski:


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -189,8 +192,10 @@ object PartitionFlow {
 
       clock <- Clock[F].instant
       timestamp <- timestamp updateAndGet (_.copy(clock = clock))
+
       states <- cache.values
-      _ <- states.values.toList.parTraverse_ { state =>
+      cores <- Runtime[F].availableCores
```
Z1kkurat, line 197: No reason to do this on every batch and timer triggering. Let's make it a part of creating the instance (`init` used in `val acquire`). We could multiply it once and pass `parallelism: Int` to `def processRecords` and `def triggerTimers`

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -252,11 +253,14 @@ object PartitionFlow {
 
     val acquire: F[PartitionFlow[F]] = init as { records =>
       for {
-        _ <- NonEmptyList.fromList(records).traverse_(processRecords)
+        cores <- Runtime[F].availableCores
```
Z1kkurat, line 256: This will also be invoked on each incoming batch. The whole `records => ...` block is a SAM implementation of `PartitionFlow`. We need something like this: 
```scala
for {
  cores <- Runtime[F].availableCores
  _ <- init
} yield { records =>
  ...
}
```

PR #446, opened 2023-02-17
Replace calls of cache.values with cache.keys

edubrovski:


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -206,17 +207,11 @@ object PartitionFlow {
...
-
-      _ <- Log[F].debug(s"got ${states.size} states from cache")
-
-      stateOffsets <- states.values.toList.traverse { state =>
-        state flatMap (_.context.holding)
-      }
-
-      _ <- Log[F].debug(s"computed stateOffsets: $stateOffsets")
+      keys <- cache.keys
+      states <- keys.toVector.parTraverseFilter(cache.get)
```
t3hnar, line 211: why do you need `_.parTraverseFilter` here?

t3hnar:
```scala
cache
.keys
.foldLeft(().pure[F]) { case (result, key) => 
 cache
  .get(key)
  .flatMap {
   case Some(value) =>
     value
     .context
     .holding
     .flatMap {
      case Some(value) =>
        result.flatMap { _.min(value) }
      case None => result
     }
   case None => 
   result
  }
 }
```

t3hnar: there is also `cache.foldMap`

Z1kkurat (approved):

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -188,12 +188,13 @@ object PartitionFlow {
...
-          state.timers.set(timestamp) *>
-            state.timers.trigger(state.flow)
-        }
+
+      keys <- cache.keys
+      states <- keys.toVector.traverseFilter(cache.get)
+
+      _ <- states.parTraverse_ { state =>
+        state.timers.set(timestamp) *>
+          state.timers.trigger(state.flow)
```
t3hnar, line 197: is this a heavy operation?

Z1kkurat: Yes, it can include sending snapshots to Kafka, Cassandra etc.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -206,17 +208,20 @@ object PartitionFlow {
...
+          case (x @ Some(_), _)   => x
+          case (_, y)             => y
+        }
+      )
 
-      minimumOffset = stateOffsets.flatten.minimumOption
+      minimumOffset <- cache.foldMap {
+        case (_, Right(value)) => value.context.holding
+        case (key, Left(_)) =>
+          Async[F].raiseError(new Exception(s"trying to compute offset to commit but value for key $key is not ready"))
```
t3hnar, line 223: I'm not sure this is safe to throw exceptions for `loading` elements

edubrovski: We should never have `Left` case here. So either
1. this never happens (most likely)
2. this happens because we have a critical bug in our code. Then I'd like to find out about this bug ASAP instead of investigating for days why kafka-flow computes incorrect offset.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -206,17 +208,20 @@ object PartitionFlow {
...
-      states <- cache.values
-
-      _ <- Log[F].debug(s"got ${states.size} states from cache")
-
-      stateOffsets <- states.values.toList.traverse { state =>
-        state flatMap (_.context.holding)
-      }
-
-      _ <- Log[F].debug(s"computed stateOffsets: $stateOffsets")
+      pickMin = CommutativeMonoid.instance[Option[Offset]](
```
Z1kkurat, line 211: This is a static instance. Let's define it in `private val` somewhere in the object, there's no need to create it every time

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
Z1kkurat: Maybe log a warning or error here if there's a need to receive a signal about such behavior? Failing the application here could prove inconvenient later because at this point the framework has already persisted snapshots and, if failed here, would restart from the previous commit but with the updated state which can be potentially incorrect, more than committing a wrong offset (but moving further nonetheless)

t3hnar: also you can just wait until it's loaded :)

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -188,12 +189,13 @@ object PartitionFlow {
...
       timestamp <- timestamp updateAndGet (_.copy(clock = clock))
-      states <- cache.values
-      _ <- states.values.toList.parTraverse_ { state =>
-        state flatMap { state =>
-          state.timers.set(timestamp) *>
-            state.timers.trigger(state.flow)
-        }
+
+      keys <- cache.keys
+      states <- keys.toVector.traverseFilter(cache.get)
```
t3hnar, line 194: could we include all in one `.parTraverse_` and not `.traverseFilter` + `parTraverse_` ?

t3hnar (approved):

PR #451, opened 2023-03-13
Update skafka and kafka-journal

edubrovski:


edubrovski:
Build is broken because there's an incompatibility in major version of skafka between kafka-flow and kafka-journal. I need to update kafka-journal.

Z1kkurat (approved):

PR #458, opened 2023-03-21
Improve logging for Kafka-based persistence

Z1kkurat:
- minor refactorings
- ensure that persistence-related log messages (`"deleted snapshot"`) are logged with a prefix (format: `"<topic>-<partition> <key>"`)

### DIFF
--- a/persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistenceModule.scala
+++ b/persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistenceModule.scala
@@ -1,14 +1,12 @@
 package com.evolutiongaming.kafka.flow.kafkapersistence
 
 import cats.Parallel
-import cats.effect.{Concurrent, Ref, Resource}
+import cats.effect.{Concurrent, Resource}
 import cats.syntax.all._
 import com.evolutiongaming.catshelper.{FromTry, Log, LogOf, Runtime}
-import com.evolutiongaming.kafka.flow.effect.CatsEffectMtlInstances._
 import com.evolutiongaming.kafka.flow.key.{Keys, KeysOf}
 import com.evolutiongaming.kafka.flow.metrics.syntax._
 import com.evolutiongaming.kafka.flow.persistence.{PersistenceOf, SnapshotPersistenceOf}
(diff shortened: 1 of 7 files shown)


persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistenceModule.scala
```diff
@@ -138,22 +136,23 @@ object KafkaPersistenceModule {
...
-              key = key,
-              database = SnapshotDatabase(
-                read  = read,
-                write = KafkaSnapshotWriteDatabase.of[F, S](snapshotTopicPartition)
-              ) withMetricsK metrics.snapshotDatabaseMetrics,
-              buffer = buffer
-            )
+        // A manual overriding of SnapshotsOf is required to pass a custom prefixed Log.
+        // Since both `SnapshotsOf.backedBy` and `Snapshots.of` are parameterized by a generic K (`KafkaKey` here),
+        // they would log the whole KafkaKey if Log.prefixed was to be called from inside.
```
theide, line 145: would it be a problem if we just used the kafka-key as log prefix? It would work with Cassandra persistence as well , right?

Z1kkurat: The only option is using `.toString` which would result in something like this: `KafkaKey(my-application-id,my-group-id,my-topic-0,kafkaKey)`. It's not the worst, but it's some unnecessary information for a prefix, especially if `applicationId` and/or `groupId` are long. Do you think it makes sense to log all of it?

theide: thanks for sharing details, I don't think .toString is a good option, but I have another idea, let's discuss it

Z1kkurat: Implemented the proposed changes, now log message looks like this in integration tests:
```
c.e.k.f.k.KafkaPersistenceModule - kafka-persistence-test-1 key1 deleted snapshot
```

theide (approved):

PR #480, opened 2023-05-16
bump pureconfig to 0.17.3

dfakhritdinov:


project/plugins.sbt
```diff
@@ -16,3 +16,5 @@ addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
 addSbtPlugin("com.evolution" % "sbt-scalac-opts-plugin" % "0.0.9")
 
 addSbtPlugin("com.evolution" % "sbt-artifactory-plugin" % "0.0.2")
+
+addDependencyTreePlugin
```
Z1kkurat, line 20: Do we need it in the project's definition? The same can be achieved locally for all project if added to `~/.sbt/1.0/plugins/plugins.sbt`

dfakhritdinov: fine, will drop it then

PR #501, opened 2023-06-19
Add option to bound parallelism for snapshot recovery

Z1kkurat:
Currently, the parallel recovery doesn't limit the number of fibers it creates when recovering the state, meaning that for each recovered key a fiber is spawned.
Source: https://github.com/evolution-gaming/kafka-flow/blob/46072b4d23e76962bb0858db86d8339d82e8633e/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala#L128-L130

This can lead to CPU exhaustion when there are too many keys and for each one a fiber is spawned, attempting to recover snapshots and consuming all CPU resources, whether it's for scheduling reasons or actual work of querying the storage.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
@@ -1,5 +1,7 @@
 package com.evolutiongaming.kafka.flow
 
+import com.evolutiongaming.kafka.flow.PartitionFlowConfig.RecoveryMode
+
 import scala.concurrent.duration._
 
 /** Configuration of `PartitionFlow`.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
```diff
@@ -21,15 +23,45 @@ import scala.concurrent.duration._
...
+
+object PartitionFlowConfig {
+
+  sealed trait RecoveryMode
+
+  object RecoveryMode {
+
+    /** Read snapshots in parallel with a limit on the number of fibers spawned. For each key a fiber will be spawned,
+      * but a total number of concurrent fibers will not exceed the specified limit. This mode is slower than
+      * `ParallelUnbounded`, but it allows fine-tuning the number of concurrent fibers to prevent CPU starvation. Just
```
theide, line 49: > but it allows fine-tuning the number of concurrent fibers to prevent CPU starvation

 but it allows fine-tuning the number of concurrent fibers to prevent CPU starvation and overwhelming the underlying storage with too many parallel recoveries.

theide (approved):

PR #545, opened 2023-09-13
Add option to limit parallelism of timers

Z1kkurat:


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -191,20 +191,21 @@ object PartitionFlow {
...
+      cache.get(key).flatMap(_.traverse_(state => state.timers.set(timestamp) >> state.timers.trigger(state.flow)))
+
     def triggerTimers = for {
       _ <- log.debug("triggering timers")
 
       clock     <- Clock[F].instant
       timestamp <- timestamp updateAndGet (_.copy(clock = clock))
 
       keys <- cache.keys
-      _ <- keys.toVector.parTraverse_ { key =>
```
t3hnar, line 201: what's expected number of `keys` here?

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
```diff
@@ -66,4 +67,27 @@ object PartitionFlowConfig {
...
+    case object Unbounded extends TimersExecutionMode
+
+    /** Execute timers in parallel with a limit on the number of concurrent executions. For each timer (corresponds to a
+      * key) a fiber will be spawned, but a total number of concurrent fibers will not exceed the specified limit. This
+      * mode is slower than `Unbounded`, but it allows fine-tuning the number of concurrent fibers to prevent CPU
+      * starvation and overwhelming the underlying storage with too many parallel executions.
+      * @param parallelism
+      *   the upper bound on the number of concurrent fibers
+      */
+    case class Bounded(parallelism: Int) extends TimersExecutionMode
```
t3hnar, line 91: why don't we just bound to number of CPU * multiplier? hence not operate with a limit but rather a multiplier?

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
```diff
@@ -33,10 +33,11 @@ import scala.concurrent.duration._
   *   Try committing everything when partition is revoked.
   */
 case class PartitionFlowConfig(
-  triggerTimersInterval: FiniteDuration = 1.second,
-  commitOffsetsInterval: FiniteDuration = 1.minute,
-  recoveryMode: RecoveryMode            = RecoveryMode.ParallelUnbounded,
-  commitOnRevoke: Boolean               = false
+  triggerTimersInterval: FiniteDuration    = 1.second,
+  commitOffsetsInterval: FiniteDuration    = 1.minute,
+  recoveryMode: RecoveryMode               = RecoveryMode.ParallelUnbounded,
```
t3hnar, line 38: `RecoveryMode.ParallelUnbounded` & `TimersExecutionMode.Unbounded` are those the same?

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
Z1kkurat: There's no expected number, it all depends on a specific application. Some might have hundreds, some might have hundreds of thousands

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
Z1kkurat: No, they are responsible for completely different parts of the flow and should be configured separately

t3hnar: I was more thinking about reusing `ParallelizationMode` in different configs… `ParallelizationMode.Unbounded | Bounded | Sequential` etc…

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
t3hnar: JFYI: we were fixing problem of calling `parTraverse` / `parSequence` against huge collections and this is what helped:

```scala
    def parFoldMap2[F[_]: Monad: Parallel: Runtime, B: Monoid](
      f: A => F[B]
    ): F[B] = {
      if (self.isEmpty) {
        Monoid[B].empty.pure[F]
      } else {

        def fold(self: Iterable[A]) = {
          self.foldLeft(Monoid[B].empty.pure[F]) {
            case (b, a) =>
              for {
                b <- b
                a <- f(a)
              } yield {
                b.combine(a)
              }
          }
        }

        Runtime[F]
          .availableCores
          .flatMap { cpus =>
            val size = self.size / cpus.max(1)
            if (size > 0) {
              self
                .grouped(size)
                .parFoldMap1(fold)
            } else {
              fold(self)
            }
          }
      }
    }

```

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
Z1kkurat: Then it would depend on external conditions, such as the configuration of an application in k8s cluster. With the absolute number, it's quite easy to calculate an absolute max number of spawned fibers per second, e.g. in case Cassandra used as a snapshot storage; all we need to know is 1) this limit (which is applied per partition); 2) a rough estimate of Cassandra query latency; 3) the number of partitions. And that is what we need in most of our use cases of this framework.  

Someone changing `requests`/`limits` of the application externally won't affect this, and an application will still produce consistent load on Cassandra whereas having a multiplier makes one change it any time one increases or decreases the number of vCPUs. This is especially relevant for increasing a number of vCPUs as it might hurt the storage proportionally.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
t3hnar: this is for the case when the work we parallelise is mostly CPU bound

t3hnar: btw, not sure whether you checked, but `parTraverseN` is implemented via `parTraverse` + `semaphore`, hence it still produces huge number of fibers.

t3hnar: in case of non cpu bound work - I'd suggest using `cpu * multiplier`, where you configure multiplier based on the workload you have…

Z1kkurat: > parTraverseN is implemented via parTraverse + semaphore

Yes, I know about it. The number of fibers itself isn't an issue while they're inactive; it's the number of concurrently active fibers that try to execute some code; in timers' case it's user-defined `Tick` code (which we know nothing about, could be either CPU- or IO-intensive) plus `TimerFlow` code which is usually both CPU-intensive (encode snapshot as a JSON) and IO-intensive (write the result to the storage)

Z1kkurat: While this would be convenient to configure the limit based on the number of CPUs for CPU-intensive part of that (JSON encoding, for example), this may be not as good for IO-intensive part of work, namely Cassandra interactions, where I'd prefer to have a steady and a predictable load profile that doesn't depend on a number of CPUs

t3hnar: sure, you are the maintainer :)

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowConfig.scala
Z1kkurat: Yes, that would be convenient! Though, they're not the same as the new one doesn't have `Sequential`, and introducing it (and implementing the logic) just for reusability seems weird as sequential mode doesn't make much sense there 🤔 Plus, we lose the scaladoc that's very detailed and descriptive as it's tailored for that specific ADT...  
I need more input on that @edubrovski @Phill101 please

t3hnar: `Sequential` = `Bounded(1)`, right?

Z1kkurat: Semantically -- yes. But differs in implementation as it doesn't spawn any fibers at all, even waiting ones (`parTraverse` vs `traverse`)

edubrovski: I'd prefer to have separate ADTs.
1. I want kafka-flow to fail fast (compilation error / config parsing error) if a user tries to use `Sequential` for timers execution.
2. It's convenient for users to look at the source code of `PartitionFlowConfig` and see what variants are available.

t3hnar: >  But differs in implementation as it doesn't spawn any fibers at all

but does it mean that it **SHOULD** be different in implementation :)

edubrovski: Another thing we can do is to have a hierarchy:
```
case object Sequential extends ParallelizationMode
sealed trait Parallel extends ParallelizationMode

object Parallel {
  case object Unbounded extends Parallel
  case class Bounded(n) extend Parallel
}
```

And then use `ParallelizationMode.Parallel` for configuring timers. This way we'll avoid duplication and make invalid value (`Sequential`) unusable.

Z1kkurat: > but does it mean that it SHOULD be different in implementation? :)

I'd prefer not to spawn any fibers whenever it's not absolutely necessary. It looks weird to me having a "sequential" mode configured, but spawning 100_000 fibers, even though 99% of them is waiting at any given moment. I don't think such uniformity in configuration data classes is a good tradeoff for such runtime overhead (which is completely unnecessary!)

> Another thing we can do is to have a hierarchy

I like this idea, it covers both cases. If no other maintainers object, I'd go with this one

Z1kkurat: I've changed the hierarchy to a single ADT, please take a look and let me know your opinion

edubrovski (approved):

PR #571, opened 2023-11-21
remove _total suffix from compressor metrics

johanmast:
Remove _total suffix from compressor metrics, this is not allowed in Prometheus >=1.0.0

When scraping metrics with the old name, an error is returned:

```
java.lang.IllegalArgumentException: 'kafka_flow_compressor_compressed_bytes_total': Illegal metric name. The metric name must not include the '_total' suffix. Call PrometheusNaming.sanitizeMetricName(name) to avoid this error.
```

The new name is the same as the outcome of `PrometheusNaming.sanitizeMetricName("kafka_flow_compressor_compressed_bytes_total")`


Z1kkurat (approved):
LGTM, thanks!

Issue #592, opened 2024-03-25
Reducing coupling with kafka-journal


grzegorz-bielski:
As you may know from [kafka-flow's overview](https://evolution-gaming.github.io/kafka-flow/docs/overview.html), the library is tightly coupled with [kafka-journal](https://github.com/evolution-gaming/kafka-journal). But in reality, only the sources from the `com.evolutiongaming.kafka.flow.journal` package are involved in reading the journal.

`kafka-flow` is already used by many internal Evo apps to a great success, and a lot of them are reading generic kafka topics instead of`kafka-journal`'s journal.

Direct dependency on `kafka-journal` brings a lot of unnecessary, transient dependencies to classpath like akka, making the library much harder to use than it should be, and the path for migration to Scala 3 unclear. In many cases `kafka-journal` is just treated as a source of util classes that are not strictly related to the `kafka-journal` domain.

My suggestion would be to move code that is shared between `kafka-journal` and `kafka-flow` into separate libraries and modules. 

In particular, we can:

**1. Ensure binary and source compatibility is checked:** 
  - [x] in `kafka-flow`: https://github.com/evolution-gaming/kafka-flow/issues/593
  - [x] in `scassandra`: https://github.com/evolution-gaming/scassandra/issues/250
  - [x] in `skafka`: https://github.com/evolution-gaming/skafka/issues/420

**2. `kafka` related**
  - [x] Use `ConsumerRecord` and `ConsumerRecords` from skafka directly instead of `ConsRecord` and `ConsRecords` aliases
  - [x] Extract to `skafka` / remove / internalize:
    - [x] `PartitionOffset`
    - [x] `KafkaHealthCheck`
    - [x] `SkafkaHelper`
    - [x] `HeaderToTuple`
    - [x] `TupleToHeader`

**3. `cassandra` related**
- [x] Extract to `scassandra` / remove / internalize:
    - [x] `CassandraSession`
    - [x]  `CassandraHealthCheck`
    - [x] `Segments`
    - [x] `CassandraHelper`

**4. Extract to separate `kafka-flow-journal` and `kafka-flow-journal-cassandra`modules:**
- [ ] all `*Journal*` stuff with `com.evolutiongaming.kafka.flow.journal` from `core` and `persistence-cassandra`
[...]

PR #601, opened 2024-04-30
introduce own KeySegments instead of kafka-journal Segments

Z1kkurat:
This is a step towards https://github.com/evolution-gaming/kafka-flow/issues/592. 

- remove methods that were marked for removal since long ago
- introduce our own version of `KeySegment` instead of `Segment` from `kafka-journal` and a corresponding set of methods
- make current methods delegate to the new ones

Current methods aren't deprecated yet because there's no alternative of `CassandraKeys`, `CassandraSnapshots`, `CassandraJournals`, `CassandraPersistence` that would use `CassandraSession` from `scassandra` instead of `kafka-journal` one. I plan to gradually introduce these and mark the old implementations as deprecated once all new ones are done. Until then, I will only introduce new APIs and change the old ones to delegate to them.

Next step is introducing our own `SegmentNr`

### DIFF
--- a/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/KeySegments.scala
+++ b/persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/KeySegments.scala
@@ -1,0 +1,64 @@
+package com.evolutiongaming.kafka.flow.key
+
+import cats.Show
+import cats.kernel.{Eq, Order}
+import cats.syntax.all._
+
+/** The maximum number of segments in Cassandra table.
+  *
+  * When [[KeySegments]] is used then the segment column value is determined by consistent hashing of the key column.
+  * I.e. there always no more than [[KeySegments#value]] different values.
+  *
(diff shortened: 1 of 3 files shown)


persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/key/KeySegments.scala
```diff
@@ -0,0 +1,64 @@
...
+
+/** The maximum number of segments in Cassandra table.
+  *
+  * When [[KeySegments]] is used then the segment column value is determined by consistent hashing of the key column.
+  * I.e. there always no more than [[KeySegments#value]] different values.
+  *
+  * The logic itself could be found in [[SegmentNr]] class constructors (apply methods).
+  *
+  * The only place where such approach is used right now is a metajournal specified by
+  * [[SchemaConfig#metaJournalTable]]. This allows the fair distribution of the journal keys between the Cassandra
```
Z1kkurat, line 15: oops, copypasta

PR #629, opened 2024-09-10
Introduce a custom GH action to publish a new release only on git tag

Z1kkurat:
## Motivation
The current release process is cumbersome, requiring multiple manual steps in a strict sequence: first, updating the version in `version.sbt`, then manually creating a GitHub release with the same version, after which the release is triggered. This process can be simplified to improve efficiency and reduce potential errors.

## Changes in this PR
This PR introduces a new model for publishing, triggered by pushing git tags in the `v*` format. When such a tag is pushed, the release process automatically starts.

The `version.sbt` file has been removed and replaced with [sbt-dynver](https://github.com/sbt/sbt-dynver), which dynamically sets the version property based on the latest git tag. This ensures that when a new `vX.Y.Z` tag is pushed, the version used during publishing step is correctly set, streamlining the release process.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,7 @@ jobs:
           echo "password=${{secrets.JFROG_ACCESS_TOKEN}}" >> ${{ env.SBT_CREDENTIALS }}
 
       - name: check, test and package
-        run: sbt "clean; +show version; +all test package"
+        run: sbt "clean; versionPolicyCheck; +all test package"
 
       - name: publish on JFrog Artifactory
         run: sbt "+publish"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,53 @@
-name: Publish Release
+name: Test and publish a new release
   
 on:
-  release:
-    types: [published]
-    branches: [master]
+  push:
+    tags:
+      - 'v*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  SBT_CREDENTIALS: '.credentials'
   
 jobs:
-  release:
-    uses: evolution-gaming/scala-github-actions/.github/workflows/release.yml@v1
-    secrets: inherit
+  publish-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: coursier/cache-action@v6
+      
+      - name: setup Java and SBT
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'oracle'
+          cache: 'sbt'
+
+      - name: setup credentials
+        run: |
+          echo "realm=Artifactory Realm" >> ${{ env.SBT_CREDENTIALS }}
+          echo "host=evolution.jfrog.io" >> ${{ env.SBT_CREDENTIALS }}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -1,11 +1,53 @@
-name: Publish Release
+name: Test and publish a new release
   
 on:
-  release:
-    types: [published]
-    branches: [master]
+  push:
+    tags:
+      - 'v*'
```
mr-git, line 6: shouldn't there be restriction to branch name too? like what will happen, if tag is pushed on non-master branch?

.github/workflows/release.yml
```diff
@@ -1,11 +1,53 @@
...
+
+      - name: setup credentials
+        run: |
+          echo "realm=Artifactory Realm" >> ${{ env.SBT_CREDENTIALS }}
+          echo "host=evolution.jfrog.io" >> ${{ env.SBT_CREDENTIALS }}
+          echo "user=github-publish" >> ${{ env.SBT_CREDENTIALS }}
+          echo "password=${{secrets.JFROG_ACCESS_TOKEN}}" >> ${{ env.SBT_CREDENTIALS }}
+
+      - name: check, test and package
+        run: sbt "clean; +show version; +all test package"
```
mr-git, line 42:
```suggestion
        run: sbt "clean; versionPolicyCheck; +all test package"
```

.github/workflows/release.yml
```diff
@@ -1,11 +1,53 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  SBT_CREDENTIALS: '.credentials'
   
 jobs:
-  release:
-    uses: evolution-gaming/scala-github-actions/.github/workflows/release.yml@v1
```
mr-git, line 10: would you suggest/consider to move all our OSS libraries to this approach by creating `release.yml@v2` action?

.github/workflows/release.yml
Z1kkurat: i assume it would checkout, build and release that tag. do you maybe know if that's the case? if so then it's even better at first glance since we could build older CE2 branches if it's necessary for some reason

.github/workflows/release.yml
Z1kkurat: sure, if it works well and i find no issues with it then i don't see why not

.github/workflows/release.yml
mr-git: I believe that there is a way to allow workflow trigger for only limited set of branches, something like:
```yaml
    branches: [master, series-0.x]
```

Z1kkurat: not sure it's possible without some workarounds due to the nature of tags, see https://github.com/orgs/community/discussions/13226.  
it gets even more complicated if we move it to a generic action in `evolution-gaming/scala-github-actions` since different repos have their own legacy branches with different names so they'd have to work around that

mr-git: we can pass arguments to action

Z1kkurat: yes, though i'm not convinced yet it's a good tradeoff to implement quite a workaround (see [this stackoverflow answer](https://stackoverflow.com/questions/63745613/how-to-get-a-branch-name-on-github-action-when-push-on-a-tag/77727776#77727776)) to just ensure the tag exists on a specific branch

what are the exact benefits we're supposed to receive by introducing this complexity?

mr-git: not sure, but technically we, as Software Engineers, can try to do that 😄

PR #636, opened 2024-09-30
Make keys DDL use LeveledCompactionStrategy by default

Z1kkurat:
This fits the use case as `keys` table is read intensively during the bootstrap phase when eager recovery is used, and then it does a lot of updates/deletes which plays well with this strategy and allows avoiding putting more pressure on the compaction process. We've observed significant performance gains when using this strategy compared to the default one in a keyspace with hundreds of thousands of keys -- compaction manages to keep up and allows for a fast keys recovery


nviliunov-evolution-throwaway:
lgtm

Issue #256, opened 2021-09-23
Improve graceful shutdown sequence


nikitapecasa:
Currently it persists the state without updating committable offsets, so after shutdown we have updated state in persistent storage but no committed offsets, which is not a critical issue by itself as any sane application should support deduplication of input messages, meaning that when we start an app again it would load latest state and re-process some messages by simply ignoring them.

Improved graceful shutdown sequence should look like
- finish processing of current messages we got from `consumer.poll`
- persist the state
- collect committable offsets after state is persisted (currently `TimerFlowOf`'s release does nothing related to offsets on release)

with following conditions
- if messages' processing fails, no persistence and no commit is required
- if state's persisting fails, committable offsets cannot be updated and no commit is required

A minimum set of tests needs to be added
- if state is persisted, correct offsets are committed, so that on next app run there's no re-processing of messages
- if state's persistence failed, then commit should not happen

Chances are quite high that it won't be an easy task as
- we don't have access to pending offsets to be committed in `TimerFlowOf`)
- shutdown sequence is a result of a composition of different resources, so multiple files have to be adjusted, mb even some considerable part of the processing flow and the way we assemble the structure of Topic/Partition/Key/Timer flows

tobiajo:
Transactional mode (#833) solves this for its own mode: the offset commit is part of the snapshot producer transaction, and also committed on release (revoke and shutdown) when `commitOnRevoke` is set.

Default consumer path is unchanged, still discards the scheduled offset on rebalance revoke, so #256 stays open for it.

PR #641, opened 2024-10-14
Commit offsets on partition flow release

i-surkov:
Addresses #256 by clearing keys cache during the partition flow release, that way giving the timer flows the opportunity to flush persistence and release held offsets _before_ partition flow tries to perform the final commit.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
@@ -229,15 +229,19 @@ object PartitionFlow {
       _ <- log.debug("done triggering timers")
     } yield ()
 
-    def offsetToCommit: F[Option[Offset]] = for {
-      _ <- log.debug("computing offset to commit")
-
-      // find minimum offset if any
-      minimumOffset <- cache.foldMap[Option[Offset]] {
+    def offsetToCommitFromKeys: F[Option[Offset]] = {
+      offsetToCommit(cache.foldMap[Option[Offset]] {
         case (_, Right(value)) => value.context.holding
         case (key, Left(_)) =>
           log.error(s"trying to compute offset to commit but value for key $key is not ready").as(none[Offset])
-      }(pickMinOffset)
+      }(pickMinOffset))
+    }
+
+    def offsetToCommit(getMinOffset: F[Option[Offset]]): F[Option[Offset]] = for {
+      _ <- log.debug("computing offset to commit")
+
+      // find minimum offset if any
+      minimumOffset <- getMinOffset
 
       // maximum offset to commit is the offset of last record
       timestamp    <- timestamp.get
@@ -272,7 +276,7 @@ object PartitionFlow {
 
         clock           <- Clock[F].instant
         commitOffsetsAt <- commitOffsetsAt.get
-        offsetToCommit  <- if (clock isAfter commitOffsetsAt) offsetToCommit else none[Offset].pure[F]
+        offsetToCommit  <- if (clock isAfter commitOffsetsAt) offsetToCommitFromKeys else none[Offset].pure[F]
 
         _ <- log.debug(s"offset to commit: ${offsetToCommit.show}")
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -283,9 +287,31 @@ object PartitionFlow {
     }
 
     val release: F[Unit] = if (config.commitOnRevoke) {
-      offsetToCommit.flatMap { offset =>
-        offset.traverse_ { offset =>
-          log.info(s"committing on revoke: $offset") *> scheduleCommit.schedule(offset)
+
+      // Outside F is about collecting handles to the offsets held by existing cache entries;
+      // Inside F is about calling and aggregating those handles, which can be done even after cache was cleared;
+      def offsetsHeldByCurrentKeys: F[F[Option[Offset]]] =
```
Z1kkurat, line 293: please correct me if i'm wrong but does it make any difference if we calculate the min offset right away without having `F[F]`? both arms of the `traverse` below end up in `pure` and even if the value isn't ready yet it will be detected during the execution of the outer `F`, not the inner one. Is there any other point of making this effect nested and delaying executing it? If not, we could calculate these offsets right away and pass them with `.pure` to `offsetToCommit`

i-surkov: > both arms of the traverse below end up in pure
Not exactly, I the first branch the outer F is `pure`, and the inner one capture the `holding` calculation, while in the second, the outer F performs logging, and the inner one is pure none, since we don't have anything to calculate.

Z1kkurat: indeed, you're right! now i wonder, what is the point of not calculating these offsets beforehand? we ignore keys that are not yet ready in the outer computation, so clearing the cache shouldn't change anything in that regard

i-surkov: Clearing the cache triggers release on `TimerFlow`s, which perform `flushOnCancel` if configured, which in turn releases the pending offset holds. Currently, we release the cache _after_ performing the commits, so `TimerFlow`s perform the persistence flush successfully, but by that time nobody cares about the fact that the offsets can be committed, so there is a de-sync between persistence and committed offsets on the next app start.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -283,9 +287,31 @@ object PartitionFlow {
...
+              case (key, Left(_)) =>
+                log
+                  .error(s"trying to compute offset to commit but value for key $key is not ready")
+                  .as(none[Offset].pure[F])
+            }
+            .map(_.foldMapM(identity)(Async[F], pickMinOffset))
+        }
+
+      // We first gather "holding" handles for existing keys, then we clear the cache, forcing the entries to flush
+      // the state and remove the offset hold if they are configured to do that,
```
Z1kkurat, line 308: let's mention that this is applicable to using `TimerFlowOf#flushOnCancel` and/or other strategies with `flushOnRevoke` to make the connection between them and this more obvious

I also think it's important to elaborate on this specific case even further, mentioning that after all handles flush the state and remove their currently held offset (there will be none), the offset to commit will correspond to the offset of the latest incoming batch of records

ideally, I would prefer it to describe both cases (using/not using `flushOnCancel`) and what offset is expected to be calculated here. i'm asking this because the relation between this code and everything happening in `TimerFlowOf` is very opaque and obscure (even though those are tightly coupled), requires building the whole mental model of code execution in one's head and confuses even me every time i spend too much time away from this code. it would greatly help everyone contributing to this code in the future

core/src/test/scala/com/evolutiongaming/kafka/flow/PartitionFlowSpec.scala
```diff
@@ -322,7 +322,77 @@ class PartitionFlowSpec extends FunSuite {
...
+        _      <- flow(f.records(key1, 100, List("event1", "event2")))
+        _      <- flow(f.records(key2, 102, List("event1", "event2")))
+        _      <- flow(f.records(key3, 104, List("event1", "event2")))
+        offset <- f.pendingOffset.get
+        // No commits expected to be scheduled yet
+        _ <- IO { assert(clue(offset).isEmpty) }
+      } yield ()
+    } >>
+      // Commit to the first offset corresponding to key2 should be scheduled during the flow release
+      f.pendingOffset.get.map(offset => assertEquals(offset, Some(Offset.unsafe(102))))
```
Z1kkurat, line 390: if i understand it correctly, this might be flaky in theory as it relies on an assumption that timer flow for `key1` will be canceled (and thus flushed) before `key2`, for which we don't have any guarantees, right? i don't think `Cache#clear` specifies any ordering guarantees

also, since we're calling `cache.clear` directly in the release function of `PartitionFlow`, how does a failure of one of the keys not get propagated there thus preventing releasing from progressing further? 🤔

i-surkov: We don't need any ordering guarantees here because `cache.clear.flatten` will finish when _all_ keys are released.
`flushOnCancel` is resource release nested in the release of a `TimerFlowOf`, which is nested in the release of `KeyFlow`, which is being released when a cache entry is evicted.

Failure to  release any of the keys (which is being tested here) doesn't affect successful release of the other keys.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
i-surkov: good point, will add an extended comment

i-surkov: done

Z1kkurat (approved):
lgtm, thanks for fixing this!

PR #675, opened 2025-02-04
Add -Xsource:3

Z1kkurat:
- add `-Xsource:3` compiler option
- fix few compiler warnings
- find+replace wildcard imports to use `*` instead of underscore


Phill101 (approved):
👍

PR #696, opened 2025-03-31
Add TTL to Cassandra persistence tables

DenisNovac:


persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/cassandra/CassandraPersistence.scala
```diff
@@ -26,10 +27,11 @@ object CassandraPersistence {
     session: scassandra.CassandraSession[F],
     sync: CassandraSync[F],
     consistencyOverrides: ConsistencyOverrides,
-    keysSegments: KeySegments
+    keysSegments: KeySegments,
+    tablesTtl: Option[FiniteDuration] = None
```
Z1kkurat, line 31: let's name it `journalsTtl` please so that it's clearer

persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/journal/JournalSchema.scala
```diff
@@ -19,14 +21,18 @@ object JournalSchema {
...
 
   def of[F[_]: Monad](
     session: CassandraSession[F],
     synchronize: CassandraSync[F],
     tableName: String,
+    ttl: Option[FiniteDuration]
   ): JournalSchema[F] = new JournalSchema[F] {
     def create: F[Unit] = synchronize("JournalSchema") {
+
+      val ttlFragment = ttl.map(d => s"WITH default_time_to_live = ${d.toSeconds}").getOrElse("")
```
Z1kkurat, line 34: should there be a leading whitespace? just in case

persistence-cassandra/src/main/scala/com/evolutiongaming/kafka/flow/cassandra/CassandraPersistence.scala
DenisNovac: The idea was to use same ttl for all three tables

Z1kkurat: I don't think that would be flexible enough to be honest, different tables probably will have different requirements in regards to TTLs

DenisNovac:
Feature implemented in https://github.com/evolution-gaming/kafka-flow/pull/700

PR #833, opened 2026-06-22
Protect Kafka snapshot writes from stale writers (transactional offset binding)

tobiajo:
Breaks out the Kafka part of #828, see it for description. Solves #732 fully for Kafka persistence storage.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
@@ -280,6 +280,8 @@ object PartitionFlow {
 
         _ <- log.debug(s"offset to commit: ${offsetToCommit.show}")
 
+        // not error-handled like the revoke path below: in transactional mode a fenced periodic commit
+        // (CommitFailedException) must crash the stale instance - that is the fencing
         _ <- offsetToCommit.traverse_(offset => scheduleCommit.schedule(offset))
 
         _ <- log.debug("done with commits")
@@ -326,7 +328,12 @@ object PartitionFlow {
       offsetsHeldByCurrentKeys.flatMap { getMinOffsets =>
         cache.clear.flatten *> offsetToCommit(getMinOffsets).flatMap { offset =>
           offset.traverse_ { offset =>
-            log.info(s"committing on revoke: $offset") *> scheduleCommit.schedule(offset)
+            log.info(s"committing on revoke: $offset") *>
+              // best-effort: a transactional ScheduleCommit can fail here (e.g. fenced), but the partition is being
+              // given away anyway (the new owner replays) and the error must not escape into the rebalance callback
+              scheduleCommit
+                .schedule(offset)
+                .handleErrorWith(e => log.warn(s"committing on revoke failed for offset $offset, ignoring: $e"))
           }
         }
       }
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
@@ -69,8 +69,8 @@ object KafkaModule {
               autoOffsetReset = AutoOffsetReset.Earliest
             )
           ) evalMap { consumer =>
-            LogOf[F].apply(Consumer.getClass) map { log =>
-              Consumer(consumer.withLogging(log))
+            LogOf[F].apply(Consumer.getClass) flatMap { log =>
+              Consumer.of[F](consumer.withLogging(log))
             }
           }
       }
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
@@ -69,7 +69,7 @@ object KafkaModule {
               autoOffsetReset = AutoOffsetReset.Earliest
             )
           ) evalMap { consumer =>
-            LogOf[F].apply(Consumer.getClass) flatMap { log =>
+            LogOf[F].apply(Consumer.getClass).flatMap { log =>
               Consumer.of[F](consumer.withLogging(log))
             }
           }
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
@@ -4,6 +4,11 @@ import cats.Applicative
 import cats.effect.Ref
 import com.evolutiongaming.skafka.{Offset, OffsetAndMetadata, Partition, TopicPartition}
 
+/** Schedules an input offset to be committed for a partition. `schedule` is not necessarily a cheap in-memory record:
+  * the default implementations only update a `Ref` and never fail, but the transactional snapshot mode runs it as a
+  * Kafka transaction that performs I/O and can fail (e.g. fenced by a stale generation), so treat it as best-effort on
+  * the revoke path.
+  */
 trait ScheduleCommit[F[_]] {
 
   /** Request ("schedule") an offset to be committed for a partition during the next commit attempt */
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
@@ -6,8 +6,9 @@ import com.evolutiongaming.skafka.{Offset, OffsetAndMetadata, Partition, TopicPa
 
 /** Schedules an input offset to be committed for a partition. `schedule` is not necessarily a cheap in-memory record:
   * the default implementations only update a `Ref` and never fail, but the transactional snapshot mode runs it as a
-  * Kafka transaction that performs I/O and can fail (e.g. fenced by a stale generation), so treat it as best-effort on
-  * the revoke path.
+  * Kafka transaction that performs I/O and can fail (e.g. fenced by a stale generation). On revoke such a failure is
+  * best-effort (swallowed, since the partition is being handed off anyway); on the periodic path it instead surfaces,
+  * failing the stale instance.
   */
 trait ScheduleCommit[F[_]] {
 
--- a/docs/kafka-single-writer-design.md
+++ b/docs/kafka-single-writer-design.md
@@ -1,0 +1,184 @@
+---
+id: kafka-single-writer-design
+title: Kafka single-writer design
+sidebar_label: Kafka single-writer design
+---
+
+Design notes for the transactional snapshot mode of `kafka-flow-persistence-kafka`
+(`KafkaPersistenceModuleOf.cachingTransactional`). User-facing guarantees, costs and rollout
+guidance are in [Persistence](persistence.md#protecting-against-stale-snapshot-writes); this page
+records the mechanism and the measurements behind it.
+
+## Problem
+
+[kafka-flow#732](https://github.com/evolution-gaming/kafka-flow/issues/732): consumer-group
+ownership of the input topic does not extend to the snapshot topic. During a rebalance a previous
+owner that has not yet observed the revocation (network issue, GC pause, slow poll loop) keeps
+writing snapshots alongside the new owner. The snapshot topic is compacted (last-write-wins), so a
+stale snapshot can overwrite a newer one; the next recovery then loads stale state and loses the
+events between the two snapshots, even though their input offsets were committed. Overlaps of tens
+of seconds have been observed in production.
+
+```mermaid
+sequenceDiagram
+    participant A as Owner A<br/>(previous)
+    participant ST as Snapshot topic<br/>(compacted: last-write-wins)
+    participant B as Owner B<br/>(new)
+
+    Note over A: folds input up to offset 100,<br/>state buffered, not yet flushed
+    Note over A,B: rebalance — partition revoked from A and assigned to B,<br/>but A has not observed it yet
+    B->>ST: recover: read to end (no newer snapshot)
+    B->>B: fold input up to offset 150
+    B->>ST: write snapshot @150 ✓
+    A-->>ST: flush buffered snapshot @100<br/>(stale: A no longer owns the partition)
+    Note over ST: last write wins — @100 overwrites @150
+    Note over ST,B: next recovery loads @100<br/>(events 101 to 150 lost)
+```
+
+## Mechanism: generation fencing
+
+The input-offset commit moves out of the consumer and **into the snapshot producer's transaction**
+via `sendOffsetsToTransaction(offsets, consumerGroupMetadata)` (KIP-447): the group coordinator
+validates the consumer **generation** and rejects a commit from a stale one (`ILLEGAL_GENERATION`).
+Since that commit and the snapshot writes share a transaction, the rejection aborts the writes too.
+The generation — authoritative for partition ownership — gates both, so a stale owner can neither
+advance offsets nor overwrite a newer snapshot.
+
+```mermaid
+sequenceDiagram
+    participant B as Stale owner<br/>(old generation)
+    participant TC as Broker<br/>(txn + group coordinator)
+    participant ST as Snapshot topic
+
+    B->>TC: beginTransaction
+    B->>ST: send stale snapshot (buffered in txn)
+    B->>TC: sendOffsetsToTransaction(offset, gen=old)
+    TC-->>B: ILLEGAL_GENERATION (partition reassigned)
+    B->>TC: abort — stale snapshot never committed
+    Note over ST: newer snapshot survives
+```
+
+This is corruption prevention, not exactly-once: output produces stay outside the transaction, so
+output is at-least-once (see [Persistence](persistence.md#protecting-against-stale-snapshot-writes) non-goals). The
+committed offset is the minimum held offset, always behind durable state, so it can never outrun the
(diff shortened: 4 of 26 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
```diff
@@ -69,8 +69,8 @@ object KafkaModule {
               autoOffsetReset = AutoOffsetReset.Earliest
             )
           ) evalMap { consumer =>
-            LogOf[F].apply(Consumer.getClass) map { log =>
-              Consumer(consumer.withLogging(log))
+            LogOf[F].apply(Consumer.getClass) flatMap { log =>
```
Z1kkurat: let's get rid of dotless invocation please since we're modifying this line already

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
```diff
@@ -4,6 +4,11 @@ import cats.Applicative
 import cats.effect.Ref
 import com.evolutiongaming.skafka.{Offset, OffsetAndMetadata, Partition, TopicPartition}
 
+/** Schedules an input offset to be committed for a partition. `schedule` is not necessarily a cheap in-memory record:
+  * the default implementations only update a `Ref` and never fail, but the transactional snapshot mode runs it as a
+  * Kafka transaction that performs I/O and can fail (e.g. fenced by a stale generation), so treat it as best-effort on
+  * the revoke path.
```
Z1kkurat: I see there's a more detailed explanation of why it's best-effort only on the revoke path in the comments to `PartitionFlow`. Should we reference those comments here somehow so that one can go and read the motivation there?

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+[kafka-flow#732](https://github.com/evolution-gaming/kafka-flow/issues/732): consumer-group
+ownership of the input topic does not extend to the snapshot topic. During a rebalance a previous
+owner that has not yet observed the revocation (network issue, GC pause, slow poll loop) keeps
+writing snapshots alongside the new owner. The snapshot topic is compacted (last-write-wins), so a
+stale snapshot can overwrite a newer one; the next recovery then loads stale state and loses the
+events between the two snapshots, even though their input offsets were committed. Overlaps of tens
+of seconds have been observed in production.
+
+```mermaid
```
Z1kkurat, line 20: should we also add a mention of the offsets being committed by owner B after saving snapshots to the diagram? otherwise it might be unclear why events 101 to 150 are "lost" (lost to what/why)

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+    B->>ST: write snapshot @150 ✓
+    A-->>ST: flush buffered snapshot @100<br/>(stale: A no longer owns the partition)
+    Note over ST: last write wins — @100 overwrites @150
+    Note over ST,B: next recovery loads @100<br/>(events 101 to 150 lost)
+```
+
+## Mechanism: generation fencing
+
+The input-offset commit moves out of the consumer and **into the snapshot producer's transaction**
+via `sendOffsetsToTransaction(offsets, consumerGroupMetadata)` (KIP-447): the group coordinator
```
Z1kkurat: shall we add a link to the KIP for clarity?

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+    B->>B: fold input up to offset 150
+    B->>ST: write snapshot @150 ✓
+    A-->>ST: flush buffered snapshot @100<br/>(stale: A no longer owns the partition)
+    Note over ST: last write wins — @100 overwrites @150
+    Note over ST,B: next recovery loads @100<br/>(events 101 to 150 lost)
+```
+
+## Mechanism: generation fencing
+
+The input-offset commit moves out of the consumer and **into the snapshot producer's transaction**
```
Z1kkurat: > The input-offset commit moves out of the consumer

I think we might want to clarify this. Without the context of knowing that actual committing happens in scope of high-level abstraction (namely, `TopicFlow`) I can see how the current phrasing might be confusing

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+    B->>TC: beginTransaction
+    B->>ST: send stale snapshot (buffered in txn)
+    B->>TC: sendOffsetsToTransaction(offset, gen=old)
+    TC-->>B: ILLEGAL_GENERATION (partition reassigned)
+    B->>TC: abort — stale snapshot never committed
+    Note over ST: newer snapshot survives
+```
+
+This is corruption prevention, not exactly-once: output produces stay outside the transaction, so
```
Z1kkurat: > output produces stay outside the transaction

I assume this is because the transaction is started strictly after the user code is executed. We might want to clarify that, otherwise might be unclear why output produces (if any in user code) stay outside the transaction

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+output is at-least-once (see [Persistence](persistence.md#protecting-against-stale-snapshot-writes) non-goals). The
+committed offset is the minimum held offset, always behind durable state, so it can never outrun the
+snapshots on disk even though offset and writes may land in different transactions.
+
+Key points:
+
+- **Every** transaction carries the partition's committable offset, so every write is gated. A
+  `ScheduleCommit` forces an offset-only transaction so progress and the on-revoke offset commit
```
Z1kkurat: by "progress" do you mean regular periodic commits here or is it something else?

> so progress and the on-revoke offset commit even with no writes pending

So, the factual behaviour is preserved (regular commits and on-revoke commits happen) but they're done using `consumer.sendOffsetsToTransaction ` instead of `consumer.commit` according to the last point below, is that correct? And it now happens directly inside of `PartitionFlow#apply` when calling `ScheduleCommit#schedule` instead of later in `TopicFlow`, correct?

We might want to clarify that the factual observable behaviour stays unchanged w.r.t. periodic and on-revoke commits, I think it might be unclear from the current wording ("even with no writes pending")

docs/kafka-single-writer-design.md
```diff
@@ -0,0 +1,184 @@
...
+  is not an alternative here: it suppresses rebalances only for graceful restarts within
+  `session.timeout.ms`, and does not fence a stuck owner whose session expired.)
+
+## Forward-looking
+
+[KIP-939 (participation in 2PC)](https://cwiki.apache.org/confluence/display/KAFKA/KIP-939:+Support+Participation+in+2PC)
+is the route to extend this fence to non-Kafka snapshot stores: a transactional producer in an
+externally-coordinated two-phase commit could bind a Cassandra/RDBMS snapshot write to the same
+generation-fenced Kafka offset commit, giving those backends the per-partition ownership guarantee
+this mode has — without the per-key compare-and-set the Cassandra backend uses today. Not actionable
```
tobiajo: Should be updated, incorrect now when split, and not definite what if anything will be merged for Cassandra.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaSnapshotWriteDatabase.scala
```diff
@@ -1,18 +1,183 @@
...
+            ).raiseError[F, Unit]
+        }
+      } yield ()
+  }
+
+  /** Default upper bound of snapshot writes committed in one transaction, see [[transactional]]. */
+  val DefaultMaxWritesPerTransaction: Int = 256
+
+  // a write carries its record; an offset-only marker carries None (see scheduleCommit)
+  private final case class Pending[F[_], S](
```
tobiajo, line 185: I have been back-and-forth if should be an ADT `SnapshotWrite/OffsetCommit`. Current is the result of a previous effort to keep down lines.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaSnapshotWriteDatabase.scala
```diff
@@ -1,18 +1,183 @@
...
     partitionMapper: KafkaPersistencePartitionMapper = KafkaPersistencePartitionMapper.identity,
+  ): SnapshotWriteDatabase[F, KafkaKey, S] =
+    apply(snapshotTopicPartition, partitionMapper, record => producer.send(record).flatten.void)
+
+  /** Result of [[transactional]]: the snapshot write database plus a [[ScheduleCommit]] that routes input offset
+    * commits through the same per-partition transactions as the snapshot writes.
+    */
+  final case class Transactional[F[_], S](
+    writeDatabase: SnapshotWriteDatabase[F, KafkaKey, S],
+    scheduleCommit: ScheduleCommit[F],
```
tobiajo, line 31: Needs to be blocking to drive a transaction, because there is no background committer, and `CommitFailedException` must propagate to flow and crash stale fenced instance.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPersistenceModule.scala
```diff
@@ -1,31 +1,84 @@
...
+import java.util.UUID
+
 /** A module, necessary to create a Kafka snapshot persistence.
   */
 trait KafkaPersistenceModule[F[_], S] {
   def keysOf: KeysOf[F, KafkaKey]
   def persistenceOf: SnapshotPersistenceOf[F, KafkaKey, S, ConsumerRecord[String, ByteVector]]
+
+  /** A `ScheduleCommit` that commits input offsets transactionally with the snapshot writes, when the module provides
+    * single-writer offset binding (transactional mode). `None` means offsets are committed the default way (by the
```
tobiajo: Maybe better and generic wording would be to "override the default way". Transactional does not need be mentioned.

Z1kkurat:
I was able to review it partially by now, decided to submit my current comments otherwise it'd be too many at once

docs/kafka-single-writer-design.md
tobiajo: Agree, this is a behavioral change that needs to be motivated and explained better.

docs/kafka-single-writer-design.md
tobiajo: Yes, soon added.

docs/kafka-single-writer-design.md
tobiajo: Yes, `TopicFlow.commitPending` is a no-op for the transactional flow, as commits now instead added to snapshot producer transactions.

docs/kafka-single-writer-design.md
tobiajo: The "output" here is the user's own producer output. The transaction covers only the offset commit and the snapshot write. Full transactional output (aiming for exactly-once) is out of scope.

Looking into how to clarify better.

docs/kafka-single-writer-design.md
tobiajo: "Progress" meant advancing the committed offset. Periodic commits are observably unchanged, they just go through the producer now instead of the consumer (the consumer side `commitPending` is the no-op). On revoke it does change: with `commitOnRevoke` set, the transactional mode commits on revoke through the producer, while the default consumer path discards the scheduled on-revoke offset (#256).

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
tobiajo: Decided against that, now each place is self-contained. I think referring to comments would be too fragile.

tobiajo:
~~Sorry for the noise, I tried do avoid that by working on a different branch. I have learned how to avoid that for the future.~~

Anyhow, I have tried to improve the documentation in: https://github.com/evolution-gaming/kafka-flow/pull/833/changes/2038a4f9eea071f561f53e6646b5d05da786fffa

UPDATE: filed a GitHub support ticket to remove the very many cross-references.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/KafkaModule.scala
tobiajo: Done

docs/kafka-single-writer-design.md
tobiajo: Done

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaSnapshotWriteDatabase.scala
tobiajo: Let's not, it is cleaner without.

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
Z1kkurat: that's fair. then maybe we should expand on it a bit here?

docs/kafka-single-writer-design.md
Z1kkurat: i rather meant that to someone not privy to the internal of the class hierarchy that wording might be a bit confusing - like, what does "out of the consumer" mean etc.

docs/kafka-single-writer-design.md
Z1kkurat: thanks for the explanation! i think we should try to explain that to a reader assuming they don't have an such an extensive knowledge of kafka-flow's internals as we do, otherwise i feel like this doc would be for a very handful of people

core/src/main/scala/com/evolutiongaming/kafka/flow/kafka/ScheduleCommit.scala
tobiajo: I did that a bit, I implied that when I said that each is self-contained now.

docs/kafka-single-writer-design.md
tobiajo: I think this specific doc should be allowed to go into internals as a design doc. I tried to keep separate what needed for library users in `persistence.md`. I should probably remove from the sidebar to not mistake it as is general library documentation.

 Can look into the balance between again, could be something useful for library users only here.

docs/kafka-single-writer-design.md
tobiajo: Ah right, misunderstood. This is anyhow better clarified now, without internal terms.

docs/kafka-single-writer-design.md
tobiajo: The prose now aim to explain _the mechanics_ better in the design docs prose, without using internal terms. For mapping to code I added an implementation section.

Again, audience for `persisence.md` is library users and should try cover how to use it and what to take into account. While the design doc is aimed and library authors, covering low level mechanics and implementation details.

Made another documentation overhaul in the very last commit: https://github.com/evolution-gaming/kafka-flow/pull/833/changes/d00bc7df8f02ea854bca8e8e81b866e7ddb17336

docs/kafka-single-writer-design.md
tobiajo: I consider this resolved now after docs clarification.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaSnapshotWriteDatabase.scala
tobiajo: Covered now in docs why, resolving.

tobiajo:
The last commit (https://github.com/evolution-gaming/kafka-flow/pull/833/changes/f1e89c7a9866e1cc9c32f2118ede20c72a9fdaca) is also docs/operational: it frames the write-path cap as a per-partition throughput ceiling (rate ≈ cap / transaction time) and keeps offset-only commits off that cap, so it stays pure write throughput.

core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala
```diff
@@ -280,6 +280,8 @@ object PartitionFlow {
 
         _ <- log.debug(s"offset to commit: ${offsetToCommit.show}")
 
+        // not error-handled like the revoke path below: in transactional mode a fenced periodic commit
+        // (CommitFailedException) must crash the stale instance - that is the fencing
         _ <- offsetToCommit.traverse_(offset => scheduleCommit.schedule(offset))
```
Z1kkurat, line 285: i wonder if we need to fail only on some specific subtypes of errors? in the default mode all `schedule` does is updating a `Ref`, and then in `TopicFlow` a commit error would be swallowed by error handling (which is fine as long as snapshots are saved successfully since it would only result in a longer replay from the latest commit). With the transactional mode any error resulting from a commit attempt would crash the instance (not only `CommitFailedException`) - is that intended?

Z1kkurat: ah, please disregard, I've looked in a wrong place by mistake

tobiajo:
Did a minor tightening and stricter separation of regular docs and design doc: https://github.com/evolution-gaming/kafka-flow/compare/0b8d78403a966c6f276f228858e954a154c08529..c8563e0c0ae81d8d350174b4e2a4b266a09bc839

Stepping back for real now 😀.

PR #838, opened 2026-06-27
Protect Cassandra snapshot writes on persist (compare-and-set)

tobiajo:
**Stacked on (merge first):** #854

### Problem

Consumer-group ownership of the input topic does not extend to the snapshot store (#732): a zombie that lost its partition but is still flushing can overwrite a newer snapshot, and the next recovery loads stale state while resuming from the newer committed offset. The Kafka backend closes this with transactional writes; the Cassandra backend was still last-write-wins.

### Fix

Guard the persist path with a compare-and-set on the offset (a lightweight transaction): a write lands only if it advances the stored offset, so a stale writer's flush is rejected (`SnapshotWriteConflict`) instead of clobbering the owner's state. The first write of a key uses `INSERT ... IF NOT EXISTS`, and a guard-expired row (present with a null offset after a partial write's TTL expiry) is repaired by claiming it via `IF offset = null`, so no key can conflict forever. Opt-in: `snapshotCompareAndSet` on `CassandraPersistence`, off by default.

This is the persist-only slice: `delete` stays plain last-write-wins (offset-gating it is an API break, kept for the full compare-and-set mode stacked on this). Accepted residual: a stale writer can revive a just-deleted key. No `core` change — buffer and recovery stay exactly as on master.

### Tests and docs

Store-level compare-and-set suite (`SnapshotSpec`), TTL edge cases (`SnapshotTtlEdgeSpec`), and the flow-level corruption reproduction and prevention (`FlowSpec`). New design doc, `docs/cassandra-single-writer-design.md` — including why the offset, not the consumer generation, is the fence token here; persistence docs updated.

### DIFF
--- a/docs/cassandra-single-writer-design.md
+++ b/docs/cassandra-single-writer-design.md
@@ -1,0 +1,246 @@
+---
+id: cassandra-single-writer-design
+title: "Cassandra single-writer design: persist only"
+sidebar_label: "Cassandra single-writer design: persist only"
+---
+
+Design notes for the compare-and-set snapshot mode of `kafka-flow-persistence-cassandra`
+(`CassandraSnapshots.withSchema(compareAndSet = true)`). What ships is small — one conditional
+predicate on each `persist` — so these notes spend little time on mechanics. Their weight is on the
+design choices: where the persist-only line is drawn and why, and the full solution (offset-gated
+deletes) that was designed and then deferred — because the reasons it was deferred are
+the main thing a future author needs before reopening it.
+
+Operational guidance (enabling, consistency levels, TTL, rollout) is in
+[Persistence](persistence.md). The Kafka backend solves the same problem with a different mechanism —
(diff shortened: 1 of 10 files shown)


docs/cassandra-single-writer-design.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Delete
+
+In this persist-only mode a delete is an ordinary `DELETE` (last-write-wins): the `IF offset <= :offset`
+guard protects *persists*, not deletes. A lagging zombie can therefore still resurrect a just-deleted key
+at a lower offset. Gating deletes on an offset (an offset-carrying logical tombstone) is the separate
+**safe-delete** layer stacked on top of this one — it is intentionally out of scope here, because it
+forces a source-breaking `SnapshotWriteDatabase.delete(key, offset)` change.
+
+## The replay window
```
tobiajo: This is only relevant for deletes I think. I think a complete overhaul is needed.

PR #842, opened 2026-07-02
Derive transactional input topic and consumer generation from the flow

tobiajo:
The transactional persistence required the input topic twice (once to the flow, again as `TransactionalConfig.inputTopic`, unvalidated — a mismatch silently commits offsets for the wrong topic) and the driving consumer's group metadata wired by hand.

Both now come from the flow: `KafkaPersistenceModuleOf.make` takes the assigned `TopicPartition` and `PartitionFlowOf.apply` threads the consumer's `groupMetadata` through, so `TransactionalConfig.inputTopic` and the `groupMetadata` parameter are gone. With the topic out of the config, docs now recommend plain `applicationId` as `transactionalIdPrefix`.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowOf.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowOf.scala
@@ -5,16 +5,13 @@ import cats.effect.kernel.Async
 import com.evolutiongaming.catshelper.LogOf
 import com.evolutiongaming.kafka.flow.PartitionFlow.FilterRecord
 import com.evolutiongaming.kafka.flow.kafka.ScheduleCommit
-import com.evolutiongaming.skafka.{Offset, TopicPartition}
 
 trait PartitionFlowOf[F[_]] {
 
-  /** Creates partition record handler for assigned partition */
-  def apply(
-    topicPartition: TopicPartition,
-    assignedAt: Offset,
-    scheduleCommit: ScheduleCommit[F]
-  ): Resource[F, PartitionFlow[F]]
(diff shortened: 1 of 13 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlowOf.scala
```diff
@@ -5,15 +5,24 @@ import cats.effect.kernel.Async
...
+    *   consumer has joined. It is re-evaluated on each use, so pass it along unevaluated - a memoized (evaluate-once)
+    *   value would go stale on the next rebalance. Used by the transactional Kafka snapshot persistence to fence stale
+    *   writers (KIP-447); flows that do not need it can ignore it.
+    */
   def apply(
     topicPartition: TopicPartition,
     assignedAt: Offset,
-    scheduleCommit: ScheduleCommit[F]
+    scheduleCommit: ScheduleCommit[F],
+    groupMetadata: F[Option[ConsumerGroupMetadata]]
```
tobiajo: To avoid "shoehorning" `groupMetadata` into `PartitionFlowOf` [as part of `PartitionAssignment`], where it can be argued to not belong. With the premise of generation is a kafka-persistence-only concern: a solution is to make `kafkaEagerRecovery: PartitionFlowOf` a `TopicFlowOf`.

See this alternative: https://github.com/tobiajo/kafka-flow/compare/master...claude/sketch-topicflowof-closure

The total diff is smaller, but generation propagation becomes hidden magic (which is negative I think), it also does not generalize to other backends.

Z1kkurat (approved):

PR #843, opened 2026-07-02
Refresh the captured consumer generation after every poll

tobiajo:
Stacked on #842 — merge that first.

Rebalance listeners fire only for partitions that changed hands, so with a cooperative assignor a rebalance can bump the generation without any callback on a member that keeps its partitions — its next transactional flush is then spuriously fenced.

Fix: also refresh the captured metadata after every poll (rebalances complete within a poll). The token can still lag the ownership epoch but never lead it, so the fence is unaffected. Covered by a new integration test with `CooperativeStickyAssignor`; docs scope the analysis to the classic rebalance protocol.


tobiajo:
Closing because all cross-reference spam. Will open a new later.

Issue #850, opened 2026-07-10
Transactional snapshot recovery silently under-reads committed snapshots past a crashed writer's open transaction (#833 follow-up)


tobiajo:
## The defect

`readSnapshots` bounds the recovery read by its own `read_committed` consumer's `endOffsets` — per the KafkaConsumer javadoc the **last stable offset (LSO)**: the minimum of the high watermark and the first offset of any open transaction. An open transaction withholds *everything above the LSO, committed or not*.

Under the per-assignment `transactional.id`s of #833 (`{prefix}-{partition}-{uuid8}`), a hard-crashed writer's transaction stays open for up to `transaction.timeout.ms` — nobody's `initTransactions` ever aborts it, because every producer gets a fresh id. That pins the LSO below records committed after it, so a recovery inside the window completes "successfully" while **silently missing a newer owner's committed snapshots**. With a second handover inside the window, the next owner recovers stale state yet resumes from the newer committed input offset — the corruption shape of #732, with no fencing violated.

**Replicated against a real broker** (KRaft 4.0.0): with a committed snapshot sitting above an open transaction, the bounded read returned in 85 ms without it.

The reachable window is a crash after the transaction's snapshot produces but before its `sendOffsetsToTransaction`: with the offset commit already pending, the consumer's `requireStable` OffsetFetch makes the next owner's `position()` wait the transaction out anyway. Note the existing "open transaction" IT does not exercise the open case — it reuses the crashed producer's `transactional.id`, so its second `initTransactions` aborts the transaction before the read runs.

## Two remedies (draft PRs, alternatives — pick one)

1. **Bound the read at the high watermark and wait** (branch `tobiajo:tj/recovery-read-high-watermark`): capture the target through a short-lived `read_uncommitted` consumer; the `read_committed` position cannot pass the LSO until the broker resolves the pin, so the read genuinely waits open transactions out (bounded by the producer's `transaction.timeout.ms` plus the broker's abort scan — up to ~70 s at the 1 min default, typically less since crash detection consumes most of the timeout). This is exactly Kafka Streams' restore shape, and for the same reason: KAFKA-10167 is Streams shipping this same under-read. Keeps per-assignment ids.

[...]

Issue #849, opened 2026-07-10
Snapshot recovery read hangs forever when its target outlives the log end (member silently evicted, invisible to health checks)


tobiajo:
## The hazard

`KafkaPartitionPersistence.readPartition` captures a target offset once (`endOffsets`) and polls until the position reaches it. If the log end regresses below the captured target — log truncation after an unclean leader election, i.e. the cluster lost acknowledged snapshot records — the target becomes unreachable and the loop spins forever. Both persistence modes share the loop (plain `caching` and `cachingTransactional`).

## Why it is worse than a hang

Recovery runs on the poll thread inside the rebalance callback (eager recovery on assignment). A hung recovery means `poll()` never returns, so after `max.poll.interval.ms` / the rebalance timeout (default 5 min) the consumer is **evicted from the group while the thread stays stuck**. No exception, no crash, no restart: the process is up, a process-level liveness probe answers fine — the instance has just silently stopped being a group member. The partition moves to another instance, whose recovery hangs on the same truncated topic, and the group degrades member by silent member. Only a health check wired to consumer-group membership or lag would notice.

## Proposed direction

Fail loudly after a bounded no-progress period, deliberately **below** `max.poll.interval.ms`, so a stuck recovery fails and leaves the group cleanly *before* the zombie eviction — the failure then surfaces as a flow error on the affected instance (visible to standard supervision), not as a quiet group departure. See draft PR (branch `tobiajo:tj/recovery-read-stall-tripwire`).

tobiajo:
With #857 merged I consider this issue closed.

PR #857, opened 2026-07-17
Recover transactional snapshots completely, with a bounded read

tobiajo:
Transactional snapshot recovery had two gaps, surfaced by the research linked below; both are closed here, with the mechanism documented in the included design doc.

**The silent miss ([#850](https://github.com/evolution-gaming/kafka-flow/issues/850)).** Under `read_committed` the read stopped at the last-stable-offset, so an open transaction pinned it below newer committed snapshots: recovery returned incomplete state, and a second handover resumed from the newer committed offset over the stale state — the corruption shape of [#732](https://github.com/evolution-gaming/kafka-flow/issues/732), with no fence violated. The read now targets the high watermark (captured up front through a short-lived `read_uncommitted` consumer) and waits out open transactions below it — the shape Kafka Streams' exactly-once restore settled on ([KAFKA-10167](https://issues.apache.org/jira/browse/KAFKA-10167)). A stable per-partition `transactional.id` keeps the wait sub-second (a takeover's `initTransactions` aborts a crashed predecessor's transaction); fencing deliberately stays with the consumer generation, so completeness rests on the read bound, never on naming.

**The silent hang ([#849](https://github.com/evolution-gaming/kafka-flow/issues/849)).** Truncation (the log end regressing below the captured target) or a hanging transaction ([KIP-664](https://cwiki.apache.org/confluence/display/KAFKA/KIP-664%3A+Provide+tooling+to+detect+and+abort+hanging+transactions) detects and aborts these; [KIP-890](https://cwiki.apache.org/confluence/display/KAFKA/KIP-890%3A+Transactions+Server-Side+Defense) brokers prevent them) could keep the read from its target forever — on the poll thread, so `max.poll.interval.ms` evicted the member silently. A no-progress deadline (default 3 minutes, reset by every position advance) fails the read with `RecoveryReadStalledError` instead, diagnosed against a re-read log end (truncation versus a transaction that outlived the deadline — opposite operator responses), and the restarted recovery captures a fresh target.

**Follow-up commit.** A post-merge adversarial review found the ephemeral recovery consumers inheriting the caller's `groupId` and `autoCommit` (on by default in skafka): a committed offset from a prior read pre-empts the `earliest` reset and silently truncates the next recovery. Pre-existing since the first snapshot-recovery read; both fields are now forced off at the shared chokepoint, and the module spec seeds the hazardous shape and asserts both consumers come out cleared.

**Review aid.** The recovery read under stall on one page — the poll loop's advancing/waiting turns, the four scenarios, the verbatim log lines; kept outside the PR:

Recovery-read stall cheat sheet

**Backing research.** Worked out in the [research corpus](https://github.com/tobiajo/kafka-flow/tree/tj/address-partition-ownership-overlap-possiblity-models/research) ([#835](https://github.com/evolution-gaming/kafka-flow/pull/835)) — requirements register, TLA+ models of the read bound and the stall deadline, the remedy decision, the follow-up as finding F-12. The checked-in material stands alone and does not reference it.

### DIFF
--- a/persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
+++ b/persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
@@ -1,7 +1,7 @@
 package com.evolutiongaming.kafka.flow.kafkapersistence
 
 import cats.implicits.*
-import cats.{FlatMap, Monad, data}
+import cats.{data, FlatMap, Monad, MonadThrow}
 import com.evolutiongaming.catshelper.{BracketThrowable, Log}
 import com.evolutiongaming.kafka.flow.kafka.Codecs.*
 import com.evolutiongaming.skafka.*
@@ -11,6 +11,7 @@ import com.evolutiongaming.skafka.consumer.{
   ConsumerConfig,
   ConsumerOf,
   ConsumerRecord,
+  IsolationLevel,
   WithSize
 }
 import scodec.bits.ByteVector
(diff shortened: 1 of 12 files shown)


persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
```diff
@@ -62,37 +126,82 @@ object KafkaPartitionPersistence {
...
-            targetOffset
-          )
-          _ <- Log[F].info(
-            s"Snapshot topic $snapshotTopic partition $partition read complete at offset $targetOffset, ${bytesByKey.size} keys read"
-          )
-        } yield bytesByKey
+
+    def endOffset(consumer: SkafkaConsumer[F, String, ByteVector]): F[Offset] =
+      consumer.endOffsets(snapshotPartitionSingleton).flatMap { endOffsets =>
+        BracketThrowable[F].fromOption(endOffsets.get(snapshotsPartition), MissingOffsetError(snapshotsPartition))
```
Z1kkurat, line 145: are end offsets present when the topic was just created and the application is run for the first time against it? won't this implementation cause it to inevitably fail in that scenario?

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
```diff
@@ -62,37 +126,82 @@ object KafkaPartitionPersistence {
...
       }
+
+    // the true log end through a short-lived read_uncommitted view: the read consumer's own read_committed
+    // endOffsets is the last-stable-offset, which an open transaction pins below records committed after it began -
+    // a target there silently under-reads, while the high-watermark target waits the transaction out
+    // (see the design doc's "Recovery read")
+    val highWatermark: F[Offset] =
+      consumerOf
+        .apply[String, ByteVector](
+          suffixed(s"snapshot-$partition-hw").copy(isolationLevel = IsolationLevel.ReadUncommitted)
```
Z1kkurat, line 155: just to make sure: will this affect the existing code path when transactions aren't used? will this consumer get a different offset than before in any scenarios?

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
tobiajo: You mean the `BracketThrowable[F].fromOption(endOffsets.get(snapshotsPartition), MissingOffsetError(snapshotsPartition))`? That was existing logic.

I think this is s bit weird too, but nothing new, just a little restructure to reuse this snippet.

tobiajo: Verified: a fresh topic gives end offset 0, not a missing entry. IT-test exercises exactly this, recovery reads up to 0 and starts fresh.

persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaPartitionPersistence.scala
tobiajo: It would affect if and only if, a non-transactional snapshot consumer is (unexpectedly) configured `read_committed`. Then snapshot reads would await the committed-end-offset (LSO) reaching the uncommitted-end-offset (HW) captured at read start.

Before (master), an open transaction would let the read complete early and silently skip committed snapshots above it. The stall deadline stays transactional-only though.

So arguably an improvement, and needed e.g. when rolling back transactional -> non-transactional, where you'd keep `read_committed`.

tobiajo: The default is _read uncommitted_, so enabling the awaiting of ongoing transactions would require opt-in.

When yielding `None` here, the code flow is identical to master for non-transactional:
```
   val capturedHighWatermark: F[Option[Offset]] =
      if (consumerConfig.isolationLevel == IsolationLevel.ReadCommitted) highWatermark.map(_.some)
      else none[Offset].pure[F]
```

PR #885, opened 2026-07-30
Honour fireEvery in TimerFlowOf#unloadOrphaned

Z1kkurat:
## Problem

`TimerFlowOf.unloadOrphaned` re-fires `onTimer` on *every* timer trigger once a key has been idle longer than `fireEvery`, instead of once per `fireEvery`. `PartitionFlow.triggerTimers` runs at `triggerTimersInterval` (1s by default) and walks every cached key, so an idle-but-not-yet-unloadable key gets checked once a second.

Waste, not a correctness bug: `flush` only runs on the `canUnload` branch and the re-registered instant is the same value each time, so no extra writes and no timer-state growth. But the cost scales with the number of in-memory keys.

## Root cause

`onTimer` re-registered the next check as `touchedAt + fireEvery`, where `touchedAt` is when a record was last *processed*:

```scala
processedAt <- timers.processedAt
touchedAt    = processedAt getOrElse committedAt
...
_ <- if (canUnload) { ...flush *> context.remove }
     else register(touchedAt)
```

`touchedAt` doesn't move while the key is idle, so it's a fixed instant in the past. Once the clock passes `touchedAt + fireEvery`, every re-registration lands in the past, `Timers.expire` drops it immediately (it keeps only instants strictly after now), and the timer is due again on the next trigger.

The spin window is `[lastTouched + fireEvery, lastTouched + maxIdle)`:

- **At the defaults** (`fireEvery = maxIdle = 10.minutes`) it is empty — the key becomes unloadable exactly when the check would first re-fire. Hence unnoticed.
- **Whenever `maxIdle > fireEvery`** it is real: `fireEvery = 5.minutes, maxIdle = 1.hour` spins for 55 minutes per idle key.
- Same on the `maxOffsetDifference` branch.

## Fix

Anchor the next check to `current`, as `persistPeriodically` and `persistPeriodicallyAndUnloadOrphaned` already do:

```diff
-              register(touchedAt)
+              register(current)
```

Safe because the decision inputs (`expiredAt`, `offsetDifference`) are recomputed from a freshly-read `processedAt` on every fire, so *when* the check runs doesn't change *what* it decides. `register`'s parameter is renamed `touchedAt` -> `checkedAt` so the old name stops reading as intentional.

### :warning: One behaviour change worth reviewing

**Unload latency is now bounded by `fireEvery`, not `triggerTimersInterval`.** An idle key used to be unloaded within ~1s of crossing `maxIdle` — an accident of the spin; now it waits for the next scheduled check, up to `fireEvery` later. That matches `fireEvery`'s scaladoc ("How often the check should be performed") and both sibling flows, but a large `fireEvery` against a tight `maxIdle` will hold keys in memory longer. The fourth test pins it.

## Tests

Five tests. A shared helper ticks the clock a minute at a time and counts **actual `onTimer` invocations**, by wrapping the flow before `trigger`; flush counts wouldn't show the spin, since `unloadOrphaned` only flushes on the `canUnload` branch. Four also run `persistPeriodicallyAndUnloadOrphaned` as a live reference, so the expected numbers come from the sibling flow rather than by hand.

| test | before | after |
|---|---|---|
| does not fire before `fireEvery` has elapsed | 0 :white_check_mark: | 0 :white_check_mark: |
| fires once every `fireEvery` while idle | **26** :x: | 6 :white_check_mark: |
| keeps cadence when a record arrives between checks | **22** :x: | 6 :white_check_mark: |
| unloads at first check after `maxIdle` | **removed at minute 13** :x: | minute 15 :white_check_mark: |
| fires on every trigger when `fireEvery` is zero | 5 :white_check_mark: | 5 :white_check_mark: |

The two that pass either way are deliberate guards. The *first* check is registered at `acquire`, a path this doesn't touch, so that one pins that the flow doesn't become eager on startup. `fireEvery = 0` is a documented configuration ("perfectly fine to set these parameters to zero") that 11 pre-existing tests rely on, and both anchors degenerate to the same thing there, so the guard keeps it meaning "check on every poll". The cadence test pins the semantic choice: a record processed mid-interval must not shift the cadence — the heartbeat follows the previous **check**, not the last **record**.

`core` 94/94, `journal` 6/6, `metrics` 5/5, `persistence-cassandra` 19/19, `persistence-kafka` 19/19. `scalafmtCheck` clean. `*-it-tests` not run locally (need Cassandra/Kafka).

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala
+++ b/core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala
@@ -39,8 +39,8 @@ object TimerFlowOf {
     maxIdle: FiniteDuration   = 10.minutes,
     flushOnRevoke: Boolean    = false,
   ): TimerFlowOf[F] = { (context, persistence, timers) =>
-    def register(touchedAt: Timestamp) =
-      timers.registerProcessing(touchedAt.clock plusMillis fireEvery.toMillis)
+    def register(checkedAt: Timestamp) =
+      timers.registerProcessing(checkedAt.clock plusMillis fireEvery.toMillis)
 
     val acquire = Resource.eval {
       for {
@@ -64,7 +64,7 @@ object TimerFlowOf {
                 persistence.flush *>
                 context.remove
             } else {
-              register(touchedAt)
+              register(current)
             }
         } yield ()
       }
--- a/core/src/test/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOfSpec.scala
+++ b/core/src/test/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOfSpec.scala
@@ -315,7 +315,7 @@ class TimerFlowOfSpec extends FunSuite {
 
     // Given("flows that check every five minutes")
     // When("timers are triggered once a minute for four minutes")
-    unloadingFlowsCheckingEveryFiveMinutes foreach {
+    unloadingFlowsCheckingEveryFiveMinutes.foreach {
       case (name, flowOf) =>
         // Then("onTimer never fires")
         assertEquals(firesOverMinutes(new ConstFixture, flowOf, minutes = 4).unsafeRunSync(), 0, name)
@@ -327,7 +327,7 @@ class TimerFlowOfSpec extends FunSuite {
 
     // Given("flows that check every five minutes")
     // When("timers are triggered once a minute for half an hour")
-    unloadingFlowsCheckingEveryFiveMinutes foreach {
+    unloadingFlowsCheckingEveryFiveMinutes.foreach {
       case (name, flowOf) =>
         // Then("onTimer fires at the 5th, 10th, 15th, 20th, 25th and 30th minute")
         assertEquals(firesOverMinutes(new ConstFixture, flowOf, minutes = 30).unsafeRunSync(), 6, name)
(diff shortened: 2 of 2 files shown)


core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala
```diff
@@ -64,7 +64,7 @@ object TimerFlowOf {
                 persistence.flush *>
                 context.remove
             } else {
-              register(touchedAt)
+              register(current)
```
stasimus, line 67: (default 10 min?) instead of every ~1s trigger, growing the uncommitted-offset window  
  and idle-key residency with no doc/changelog note.

stasimus (approved):

stasimus (approved):

stasimus:
Re-triggering required check after org policy rollout fix

stasimus:
Re-running policy check with updated detector

stasimus (changes requested):
Please fix comments according to AI policy

core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala
Z1kkurat: I don't think there should be a doc note here as the current doc already states for `fireEvery` - `How often the check should be performed.`. So, the doc was misleading before because with `fireEvery` < `maxIdle` (e.g. 3m and 10m) the key would be evicted immediately after `maxIdle` (see `One behaviour change worth reviewing` in the description).  
As for the changelog - I'd put it into the release notes when making a GH release

PR #907, opened 2026-08-23
[Security] Use shared dependency-graph workflow

stasimus:
Replaces the local dependency submission job with the reusable workflow from scala-github-actions (v6.4.0). Same modules-ignore and configs-ignore, but action versions (checkout v7, setup-java v5, sbt-dependency-submission v3) are now maintained in one place.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency graph automation to use a standardized workflow.
  * Preserved dependency graph updates for changes pushed to the master branch.
  * Maintained write permissions and exclusions for designated modules.

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: [ master ]
 
+permissions:
+  contents: write
+
 jobs:
   submit:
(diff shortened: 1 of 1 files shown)


stasimus:
Good catch, added. The permissions block inside the reusable workflow can only downgrade, it can't raise anything above what the caller's token has, so with read-only defaults the submit would 403. I'm patching the same into the org-wide rollout PRs and the README snippet.

mr-git:
done in https://github.com/evolution-gaming/kafka-flow/pull/908

PR #920, opened 2026-08-26
Update scassandra to 5.7.0 and drop the pins it makes redundant

Z1kkurat:
- scassandra 5.7.0 declares patched Guava, Netty and Jackson itself (evolution-gaming/scassandra#367), so pinning them here is redundant — and now harmful: ours had fallen behind (guava 32.1.3 vs 33.7.1, netty 4.1.136 vs 4.1.137) and, as build-wide `dependencyOverrides`, would downgrade rather than raise.
- Removed the guava and netty pins.
- Kept lz4 (kafka-clients brings 1.10.2) and Jackson (kafka-journal brings 2.14.3 to `journal`, which scassandra isn't on the classpath of). Jackson raised to 2.18.10 to match scassandra so the override doesn't pull `persistence-cassandra` down.
- Guava 32→33 drops `jsr305`/`checker-qual` and moves `j2objc-annotations` to 3.x, so those join `versionPolicyIgnored`. Its comment said to drop the block after the next release, but v10.2.0 is that release — retargeted.
- Verified: resolved tree shows guava 33.7.1 and all seven netty artifacts at 4.1.137 via scassandra; `sbt check` clean; tests pass on 2.13 and 3.3; IT modules compile.

## Summary by CodeRabbit

* **Updates**
  * Updated the Cassandra testing component to version 5.7.0.
  * Updated the Jackson component to version 2.18.10.
  * Refined dependency version handling for improved build consistency.
  * Simplified dependency management by removing unnecessary explicit version pins while preserving required compression and JSON components.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -8,7 +8,7 @@ object Dependencies {
   val scache            = "com.evolution"       %% "scache"              % "6.0.2"
   val skafka            = "com.evolutiongaming" %% "skafka"              % "21.0.3"
   val sstream           = "com.evolutiongaming" %% "sstream"             % "1.3.0"
-  val scassandra        = "com.evolutiongaming" %% "scassandra"          % "5.6.0"
+  val scassandra        = "com.evolutiongaming" %% "scassandra"          % "5.7.0"
   val cassandraSync     = "com.evolutiongaming" %% "cassandra-sync"      % "4.0.0"
   val random            = "com.evolution"       %% "random"              % "1.0.5"
   val retry             = "com.evolutiongaming" %% "retry"               % "3.1.0"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -19,33 +19,24 @@ object Dependencies {
     * These are declared as direct dependencies of the modules that pull them in, on top of being listed in
     * `dependencyOverrides`: overrides only affect the resolution of this build and are not published to the POM, so on
     * their own they leave the consumers of the library with the original, vulnerable versions.
+    *
+    * Pin only what a module does not already get pinned upstream: `scassandra` covers `persistence-cassandra` since
+    * 5.7.0, and a pin lagging upstream downgrades a patched version instead of raising an unpatched one.
```
mr-git: I have a feeling, that some words are missing in this text or maybe there are some unnecessary words... not sure, but I cannot grasp the meaning :/

mr-git (approved):

project/Dependencies.scala
Z1kkurat: fixed, it was correct but indeed a bit wordy

mr-git (approved):

PR #935, opened 2026-09-08
Tolerate a stale-generation fence instead of failing the flow [DEFERRED]

### LINKED ISSUE #938: Transactional snapshot mode fails the flow on a spurious generation fence under `CooperativeStickyAssignor`
kafka-flow `master` at 1a062dc, kafka-clients 4.3.1.

## Summary

This affects `CooperativeStickyAssignor` on the classic protocol, where retained partitions keep writing through a rebalance. Eager assignors (`StickyAssignor`, `RangeAssignor`) are not affected: they tear every flow down before the generation bumps.

In `cachingTransactional` mode every snapshot flush and periodic offset commit is a Kafka transaction bound to the consumer's group generation (KIP-447). When the broker rejects one for a stale generation, kafka-flow fails the flow. The rejection already is the fence: the transaction aborted, nothing landed, the producer stays usable, and the consumer refreshes its generation on the next completed rebalance. Failing the flow adds a leave and a rejoin that bump the generation again and fence the peers' in-flight transactions. Under a rolling deploy this becomes a self-sustaining restart cascade.

## Why a still-valid owner gets fenced

The coordinator checks member and generation, not partitions (`ClassicGroup.validateOffsetCommit`). The generation advances on every completed rebalance, even one that changes nothing for this member. kafka-flow reads it after each poll, so a transaction started between the new generation forming and the next post-poll refresh carries the old one and is rejected. Under the cooperative assignor retained partitions keep committing during a rebalance, so this window is hit routinely. The design doc names it: "the retained partition's next transactional commit would be spuriously fenced though the member still owns it".

## What kafka-flow does with the rejection

`GroupCommit.commitBatch` aborts and fails every pending item ([KafkaSnapshotWriteDatabase.scala#L135-L158](https://github.com/evolution-gaming/kafka-flow/blob/1a062dc/persistence-kafka/src/main/scala/com/evolutiongaming/kafka/flow/kafkapersistence/KafkaSnapshotWriteDatabase.scala#L135-L158)). Both callers fail the flow:

- **Persist path:** `attemptToPersist` re-raises unless `ignorePersistErrors` is set ([TimerFlowOf.scala#L233-L246](https://github.com/evolution-gaming/kafka-flow/blob/1a062dc/core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala#L233-L246)); `unloadOrphaned` fails regardless of the flag ([TimerFlowOf.scala#L60-L66](https://github.com/evolution-gaming/kafka-flow/blob/1a062dc/core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala#L60-L66)).
- **Periodic commit path:** `PartitionFlow` schedules the commit without error handling ([PartitionFlow.scala#L283-L285](https://github.com/evolution-gaming/kafka-flow/blob/1a062dc/core/src/main/scala/com/evolutiongaming/kafka/flow/PartitionFlow.scala#L283-L285)).

## Why failing is redundant

The producer survives (`commitBatch` already aborts). The consumer heals on its own: a fenced member completes the rebalance round on its next `poll()` and keeps its retained partitions. Nothing was written: the key's state is still in memory and the next tick flushes it under the refreshed generation.

## How the failure amplifies

Closing the consumer sends `LeaveGroup`, which opens a rebalance and bumps the generation, fencing every peer with a transaction in flight. The retry joins, usually a second rebalance and a second bump. Each fenced peer repeats the cycle. Every restart also re-runs partition recovery, so the cost grows with the snapshot topic.

## The existing escape and its defect

`ignorePersistErrors` keeps the flow alive through a fenced flush but tolerates every persist error, so a genuine fault stalls the committed offset silently. It covers neither `unloadOrphaned` nor the periodic commit. In `persistPeriodicallyAndUnloadOrphaned` ([TimerFlowOf.scala#L189-L198](https://github.com/evolution-gaming/kafka-flow/blob/1a062dc/core/src/main/scala/com/evolutiongaming/kafka/flow/timer/TimerFlowOf.scala#L189-L198)) it also removes the key whether or not its persist succeeded, so the committed offset can advance past state that was never persisted.

## Related

- #732 describes the stale-writer overlap that transactional mode fences. The fence holds; this is about what the flow does after.
- Fixes: #935 tolerates the fence on both paths; #934 fixes the unload defect independently.

tobiajo:
In transactional snapshot mode (`cachingTransactional`) a stale-generation rejection of the offset commit (`ILLEGAL_GENERATION`, surfaced as `CommitFailedException`) failed the flow, on both the persist path and the periodic commit path. The rejection already is the fence: the transaction aborted, nothing landed, the producer stays usable after the abort, and the consumer refreshes its generation once it completes the rebalance. Failing the flow only adds a leave and a rejoin, each of which bumps the generation and fences the peers' in-flight transactions; under `CooperativeStickyAssignor` retained partitions keep committing while a rebalance completes, so a rolling deploy becomes a restart cascade.

- `KafkaSnapshotWriteDatabase.GroupCommit.commitBatch` classifies the rejection once, as the new `GenerationFencedError` (a `core` type; `core` still imports nothing from kafka-clients).
- Persist path (`persistPeriodically`, `persistPeriodicallyAndUnloadOrphaned`, `AdditionalStatePersist`): warn, the key stays dirty and keeps holding its offset, regardless of `ignorePersistErrors`. The timer flow retries on its next tick; an additional persist leaves the state to the next persist. `unloadOrphaned` and the revoke-time flush are unchanged. `persistPeriodicallyAndUnloadOrphaned` must not unload a key whose persist was tolerated, so this PR also gates the unload on the persist outcome; that is the same change as #934, which fixes it for `ignorePersistErrors` on its own. Whichever merges second rebases trivially.
- Periodic commit path (`PartitionFlow`): `committedOffset` advances only after a successful schedule; a fenced commit is warned and the offset is scheduled again on the next tick, or committed sooner by the next snapshot write.
- Every other error is handled as before. Eviction (`UNKNOWN_MEMBER_ID`) surfaces as the same exception and is tolerated too: nothing lands, and the consumer's next rejoin reports it as `onPartitionsLost`, which tears the flows down.
- The docs and the integration test that described the crash as the fence now describe the retry; the integration test also shows the retried flush landing on the same producer once the generation is current again.

Fixes #938.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -178,6 +178,9 @@ lazy val `persistence-kafka-it-tests` = (project in file("persistence-kafka-it-t
   .settings(commonSettings)
   .settings(
     name := "kafka-flow-persistence-kafka-it-tests",
+    // one broker container per suite, and the rebalance suites are paced by real coordinator timeouts: run in
+    // parallel they starve each other (the churn suite took 38 minutes next to four peers, 2 alone)
+    Test / parallelExecution := false,
     libraryDependencies ++= Seq(
       catsHelperLogback            % Test,
       playJsonJsoniter             % Test,
--- a/build.sbt
+++ b/build.sbt
@@ -75,6 +75,8 @@ lazy val core = (project in file("core"))
     mimaBinaryIssueFilters ++= Seq(
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.evolutiongaming.kafka.flow.KeyFlow.of"),
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.evolutiongaming.kafka.flow.KeyFlowOf.apply"),
+      // `Snapshots.apply` is private to the `snapshot` package; Scala 3 still emits it as a static method
+      ProblemFilters.exclude[DirectMissingMethodProblem]("com.evolutiongaming.kafka.flow.snapshot.Snapshots.apply"),
     ),
     libraryDependencies ++= Seq(
       Cats.core,
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import Dependencies.*
+import com.typesafe.tools.mima.core.*
 
 ThisBuild / versionScheme := Some("early-semver")
 ThisBuild / evictionErrorLevel := Level.Warn
@@ -67,6 +68,14 @@ lazy val core = (project in file("core"))
   .settings(commonSettings)
   .settings(
     name := "kafka-flow",
+    // `KeyFlow.of` and `KeyFlowOf.apply` build the delete path that tolerates a stale-generation fence, which needs
+    // `MonadThrow` where `Monad` was enough. The constraint is the only change, and every effect these are used with
+    // (`Sync`, `Async`, `IO`) already satisfies it, so callers recompile unchanged; only one linked against 10.3.x
+    // without recompiling would miss the method. Drop these if the fix goes out as a major instead.
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.evolutiongaming.kafka.flow.KeyFlow.of"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.evolutiongaming.kafka.flow.KeyFlowOf.apply"),
+    ),
     libraryDependencies ++= Seq(
       Cats.core,
       Cats.mtl,
--- a/build.sbt
+++ b/build.sbt
@@ -160,6 +160,11 @@ lazy val `persistence-kafka` = (project in file("persistence-kafka"))
   .settings(commonSettings)
   .settings(
     name := "kafka-flow-persistence-kafka",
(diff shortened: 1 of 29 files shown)


tobiajo:
Discovered an issue with this change. Will iterate on it with better testing before undrafting.

tobiajo:
An alternative to this change is to simply discourage `CooperativeStickyAssignor`, as it is the only assignor with this issue: eager assignors tear every flow down before the generation bumps, at the price of a full re-recovery of every partition on every rebalance. The `consumer` rebalance protocol ([KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol)) with the upcoming [KIP-1251](https://cwiki.apache.org/confluence/spaces/KAFKA/pages/399279344/KIP-1251+Assignment+epochs+for+consumer+groups) is expected to not have issues with spurious fencing.

PR #942, opened 2026-09-09
Fix dependabot alerts in website lockfile

stasimus:
Clears the four open dependabot alerts. js-yaml goes to 3.15.2 and 4.3.2.

qs needed a resolution in package.json, since express and body-parser both pin it to ~6.15.1 and that caps below the patched 6.16.0.

Checked that yarn install --frozen-lockfile and yarn build still pass.

## Summary by CodeRabbit

- **Chores**
  - Updated project configuration. No user-facing functionality or behavior has changed.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -22,6 +22,9 @@
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
   },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


stasimus:
website client vulnerabilities fixes, code is untouched

PR #947, opened 2026-09-22
Pin sbt to 1.x in Scala Steward

stasimus:
Scala Steward keeps proposing sbt 2.x (see #946), but the build doesnt work on sbt 2 yet: overlapping output dirs, Cassandra testcontainers, mdoc. Pinning to 1.x until we migrate.


mr-git (approved):
Probably have to introduce the issue with listed blockers for eventual upgrade.

stasimus:
#948

